PR #2, opened 2020-11-23
Jira authentication

AleksiKytola:


server/src/api/tokens/tokens.model.ts
```diff
@@ -0,0 +1,39 @@
+import { Model } from 'objection';
+import User from './../users/users.model';
+
+export default class JiraConfiguration extends Model {
```
AleksiKytola, line 4: class Token extends Model?

PR #90, opened 2021-06-23
Add component for form fields

juhoeerola:
The component wraps:
- optional label
- input element
- error message for the input

### DIFF
--- a/frontend/src/components/forms/FormField.tsx
+++ b/frontend/src/components/forms/FormField.tsx
@@ -1,0 +1,88 @@
+import React, { useState, useRef, useEffect } from 'react';
+import classNames from 'classnames';
+import css from './FormField.module.scss';
+
+const classes = classNames.bind(css);
+
+interface FieldType extends HTMLElement {
+  readonly validationMessage: string;
+  setCustomValidity(message: string): void;
+  checkValidity(): boolean;
+}
+
+export interface FieldProps<T extends FieldType> extends React.HTMLProps<T> {
+  label?: string;
+  error?: {
+    message: string;
+    setMessage: (_: string) => void;
+  };
+}
+
+const FieldError = ({ msg }: { msg: string }) =>
+  !msg ? null : (
+    <p className={classes(css.fieldError)}>
+      <svg
+        width="14"
+        height="14"
+        viewBox="0 0 14 14"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+      >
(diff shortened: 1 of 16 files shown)


IidaKainu:
The error messages look great and function well
I noticed the errors activate after the form is first submitted. A lot of forms I've encountered show the errors as user is typing. Could be beneficial to ask Teemu how they are intended to work here (probably hard to demonstrate with figma)

juhoeerola:
> I noticed the errors activate after the form is first submitted

Once the specific field has been invalid once (on submit), it shows the validation status while typing (except validation from server, or submit handler of course).

I went with this approach, as it would be annoying if all the fields would nag about validation errors from the start.

Maybe validating a field when focus leaves it, would be an option.

IidaKainu:
Yeah I can see how it could be annoying. Validating the field on focus leave seems like an good option. At least it could decrease the amount of errors user encounters at the same time.

frontend/src/components/forms/FormField.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+function field<T extends FieldType>(
+  Tag: React.ComponentType<React.HTMLProps<T>>,
+): React.FC<FieldProps<T>> {
+  return ({ error, label, id, ...props }) => {
+    const [localError, setLocalError] = useState('');
+    const err = error ?? {
+      message: localError,
+      setMessage: setLocalError,
+    };
```
IidaKainu, line 45: Just out of curiosity: why both localError and err are used as localError is only used in err?

IidaKainu (changes requested):
I won't pretend to completely understand how the FormField works, but in testing it functions just as planned!

Labels could be added to milestone, task creation and task edition  -fields for consistency
Other than that, looks good 👍

frontend/src/components/modals/EditVersionModal.tsx
```diff
@@ -64,12 +66,13 @@ export const EditVersionModal: React.FC<EditVersionModalProps> = ({
             <label htmlFor="name">
               <Trans i18nKey="Milestone name" />
             </label>
```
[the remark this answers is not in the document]
juhoeerola: good catch

frontend/src/components/forms/FormField.tsx
juhoeerola: I think I had some linter error/warning about conditionally calling `useState`, but didn't get it anymore

Cleaned this up a little now

IidaKainu (approved):

PR #94, opened 2021-06-30
Fix waiting for ratings-icons

IidaKainu:
Business-user had a waiting for ratings -icon when it was not needed
(in practice, when they had not personally rated the task).
Business-user now has waiting for ratings -icons for customers they
represent and have not rated for yet.


juhoeerola (approved):
Looks good!

PR #95, opened 2021-07-01
Require customer email

IidaKainu:
Changed customer email to be required in backend and modified a few things relating to customers (test data, customer edition "confirm" button)

### DIFF
--- a/frontend/src/redux/roadmaps/types.ts
+++ b/frontend/src/redux/roadmaps/types.ts
@@ -33,6 +33,10 @@ export interface teamMemberRequest {
   type?: RoleType | null;
 }
 
+export interface CheckableUser extends RoadmapUser {
+  checked: boolean;
+}
+
 export interface RoadmapUser {
   id: number;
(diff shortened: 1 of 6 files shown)


frontend/src/redux/roadmaps/types.ts
```diff
@@ -33,6 +33,10 @@ export interface teamMemberRequest {
   type?: RoleType | null;
 }
 
+export interface CheckableUser extends RoadmapUser {
+  checked: boolean;
+}
```
juhoeerola, line 38: This definition could be used also in https://github.com/Vincit/VISDOM-Roadmapper/blob/ec59d1b6e4f645cec8be0e6284e7868c505f1a62/frontend/src/components/modals/EditCustomerModal.tsx#L33

juhoeerola (changes requested):
Looks good to merge after removing the duplicate type definition

juhoeerola (approved):

PR #96, opened 2021-07-02
Modify People-page lists

IidaKainu:
Rename th 'ID' to 'Color'.
Add contact information with mailto-links to lists. Modify RoadmapUser-interface to have email and roadmap/:id/users -routes fetch email.
Add 'Unrated tasks' to lists with sorting functions.

- Feedback on sorting functions and unratedTasksAmount-function is especially welcome. I am a bit nervous of the amount of arrays iterated through multiple times.


juhoeerola (changes requested):
Looks good! Found few possible points to consider.

The iteration amount should be fine, unless it shows signs of slowing up, I'd say it can be improved if it becomes a problem.

A supplemental data structure, like `Map` or `Set` might work for some cases.

The sorting could use decorate-sort-undecorate pattern, if the sort keys require significant computation. This should even be quite simple to add to the current sorting helpers.

IidaKainu:
I've refactored code according to your points. Well, except for:
> The sorting could use decorate-sort-undecorate pattern, if the sort keys require significant computation. This should even be quite simple to add to the current sorting helpers.

I looked into DSU-pattern and I get the idea. However I am having a bit trouble implementing it here.

Should I make a new sorting utility function similar to sorted in SortUtils? It would then take a list of sets rather than a list and comparison would be done according second member of the set? 

I guess I'm not sure what conversion (decorate-part of the pattern) should be done here too

juhoeerola:
> Should I make a new sorting utility function similar to sorted in SortUtils? It would then take a list of sets rather than a list and comparison would be done according second member of the set?

This could work, but it leaves the decorating for the caller, and so the existing sorting function would also work for this (given appropriate comparison function).

Another option would be function that takes the SortKey function (and probably comparison function for the keys), this could be something like:

```ts
const DSU = <T, K>(
  list: T[],
  key: SortKey<T, K>,
  compare?: SortComparison<K>,
  order: SortingOrders = SortingOrders.ASCENDING,
) =>
  compare === undefined
    ? list
    : list
        .map((item) => ({ key: key(item), item }))      // decorate
        .sort((a, b) => compare(a.key, b.key) * order)  // sort
        .map(({ item }) => item);                       // undecorate
```

IidaKainu:
The DSU function you wrote is really nice. Decoration of roadmapUsers required extra parameters (tasks, customers), so I ended up with the first option.

I think I got the split of taskUtils-functions correct now. I moved filtering logic from taskAwaitsRatings to isUnrated and added isRatedByUser and isRatedByCustomer to clean up repeated code.
One thing was left a bit unclear. Is taskIsUnrated-function unnecessary? Can isUnrated be used on a single task somehow?

juhoeerola:
> One thing was left a bit unclear. Is taskIsUnrated-function unnecessary? Can isUnrated be used on a single task somehow?

```ts
isUnrated(user)(task)
```

The call `isUnrated(user)` returns the predicate, which can then be used to check the task.

juhoeerola (approved):

PR #97, opened 2021-07-02
Small style fix and some refactoring

juhoeerola:
The icon was positioned incorrectly on safari, now it looks quite consistent on safari, chrome and firefox.
wonky-icon

The commit messages should be somewhat self explanatory about each refactoring.


frontend/src/components/forms/SvgButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+type SvgButton<T = {}> = React.FC<T & ButtonProps & { href?: string }>;
+
+function svgButton<E = {}>(
+  Icon: React.FC<ButtonProps>,
+  {
+    hoverColor = colors.black60,
+    className,
+    ...extraProps
+  }: ExtraProps & { [K in keyof E]?: E[K] },
+): SvgButton<Pick<E, Exclude<keyof E, keyof ExtraProps>>> {
```
IidaKainu, line 34: Which props get picked here is hard to read for me (is only hoverColor overridden by ExtraProps in Pick?), so perhaps setting the default color in scss could make this a bit simpler?
`fill: var(--icon-hover-color, $COLOR_BLACK60);`
https://developer.mozilla.org/en-US/docs/Web/CSS/var()#custom_properties_with_fallbacks_for_use_when_the_property_has_not_been_set

frontend/src/components/forms/SvgButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return ({ href, ...props }) => {
+    const icon = (
+      <Icon
+        className={classes(css.svgButtonIcon, className)}
+        style={{ '--icon-hover-color': hoverColor }}
```
IidaKainu, line 39: This is a really cool way to pass variables to scss 👍

IidaKainu (approved):
Refactoring and the style fix look good!
I've added a small suggestion below: up to you whether to implement or not

frontend/src/components/forms/SvgButton.tsx
juhoeerola: I moved the default color to css, and renamed the `extraProps` parameter to a better describing `defaults`, as it can provide defaults for any optional properties in `E`.

The `Pick` will exclude any properties in `ExtraProps` from the resulting button's props, e.i. `hoverColor` and `className`.

PR #98, opened 2021-07-07
Update customer stakes visualization

IidaKainu:
The commit message explains this feature in more detail.
Here's screenshots of some examples to get an idea of how it looks:
ex02
ex07
ex08

### DIFF
--- a/frontend/src/components/TableTaskRow.module.scss
+++ b/frontend/src/components/TableTaskRow.module.scss
@@ -26,12 +26,3 @@
 .userIcon {
   color: $COLOR_BLACK60;
 }
-
-.tooltip {
-  @extend .tooltip-base;
-  margin: 0px 0px 10px 0px;
-}
-
-.tooltipArrow {
-  color: white;
-}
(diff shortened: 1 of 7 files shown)


PatricKangasmaki:
Functionality seems to be working correctly and code looks good! 

I was wondering a bit about the sizes of these dots; is the size difference between biggest and smallest possible dots a bit too much? The smaller dot might be a bit hard to see, especially if there's someone who doesn't have the greatest eye-vision. This is just a corner case though which probably does not happen often or at all (1st customer had rating of 10, 2nd had 1). I'm not a designer though so take my design advice with a grain of salt
image

IidaKainu:
> The smaller dot might be a bit hard to see, especially if there's someone who doesn't have the greatest eye-vision.

I agree. Currently the dots faithfully visualize the data. Violet gets 90.9% of the value, yellow 9.1%. So the dot diameters are 50px * 0.909 = 45.45px and 50px * 0.091 = 4.55px
Increasing the max-diameter could combat this issue a bit. Here's how it looks with 80px max-diameter
image

juhoeerola:
Some lower limit to the dot size might also help, those can still be ordered by the actual value, even though the sizes would stay the same.

Scaling the area would also be an option

IidaKainu:
Modified the max-diameter to begin from 60px and limited the dot-size multiplier to be 0.2 at the minimum. Barely-visible dots look alright to me now

PatricKangasmaki:
> Modified the max-diameter to begin from 60px and limited the dot-size multiplier to be 0.2 at the minimum. Barely-visible dots look alright to me now

Looks good now imo :+1:

PatricKangasmaki (approved):

PR #99, opened 2021-07-07
Add tooltipIcon component and use in different pages with placeholder…

islamnahidul:
* Tooltip Component added. 
* Component is used in different pages beside the header title.
* Placeholder text is used. Needs to be updated later.


frontend/src/components/forms/TooltipIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <Tooltip
+      classes={{
+        arrow: classNames(css.tooltipArrow),
+        tooltip: classNames(css.tooltip),
+      }}
+      title={title}
+      placement="top"
+      arrow
+    >
```
PatricKangasmaki, line 19: Looking good! I've got a few suggestions for the tooltip component to make it reusable around the project:

I think this component could be called just Tooltip and it could be used in other places where tooltips are needed as well (Customers in task tables, customer stakes in milestones etc.), not just around the info icons. 

`placement="top"` could be also replaced with `placement={placement}` and pass it as parameters just like the title (maybe use top as default). There might be some cases where the placement should be somewhere other than top.

juhoeerola:
Could you rebase this on top of the current master branch

The root level node_modules directory should be removed

frontend/src/components/forms/TooltipIcon.tsx
islamnahidul: It was developed for that specific task. @material-ui/core/Tooltip is already there for other generic usage and is being used already in some other places.

PR #100, opened 2021-07-08
Handle missing tokens

juhoeerola:
- server handles errors from missing/invalid oauth tokens
- frontend task import modal handles the error, showing a message telling the user to configure the tokens

Should the modal show a button for the oauth configuration modal? This might be nice for the user.

Even nicer might be, if the oauth modal would return back the the import modal on success. Would this be worth implementing?


IidaKainu:
In terms of user flow showing the button for the oauth configuration modal (on task import error) and import modal after entering a correct oauth code would be really neat. It would save a few clicks and/or possible confusion as to what to do next.

Currently the oauth modal closes after completion. So another option could be to add a final step to the modal (like in customer creation) with instructions along the lines of 'Authorization completed! Tasks can now be imported on the Tasks-page.' (with a link to the page). Then the task import error msg could have a link to settings page too?

I also wonder if an error message should be shown when entering an invalid code for oauth? Currently the loading icon is just left spinning.

juhoeerola:
Added flow for invalid tokens in import modal:
```
bad tokens in import tasks modal
-> button to open corresponding oauth config
-> oauth modal
   - success -> show original import tasks modal
   - failure -> show error message, modal can be closed
```

I disabled the submit button in oauth modal after failed token swap, because the oauth flow requires all steps from start.
Another option might be some kind of "retry" button that would reset the modal, and start over without requiring the user to close and re-open the modal.

IidaKainu:
The flow feels really intuitive! A retry button would improve it further
I am wondering if import task modal could be shown on settings page on oauth modal success too.

image
Perhaps a more descriptive error msg could be returned on failure? "Authorization failed"? The user could be confused about whether the problem is on server side or on them (bad auth code)

juhoeerola (commented):
I also reworked the modals a bit:

There are now three steps to adding a new modal, pointed in the comments. The nice property is that after the step 1/3, it won't compile before the other steps are done, so forgetting one should not be a problem.

juhoeerola:
> A retry button would improve it further

Added!  I used the cancel button style, as the normal "success" button would be misleading. Although, I think this would look nicer with different style from the cancel button.

> I am wondering if import task modal could be shown on settings page on oauth modal success too.

I don't think this would be as common use pattern, and maybe even surprising

> Perhaps a more descriptive error msg could be returned on failure? 

Showing good informative error messages would require more handling on the server side as well, now the error "type" will always be "Internal server error" as far as I could see.
I now show the error message from the server, if available.

frontend/src/components/modals/IntegrationOauthModal.tsx
```diff
@@ -106,75 +109,74 @@ export const OauthModal: React.FC<OauthModalProps> = ({
...
-              </button>
-            ) : null}
-          </ModalFooterButtonDiv>
-        </ModalFooter>
-      </Form>
-    </>
+            this URL
+          </a>{' '}
+          and input the code below:
+        </p>
```
IidaKainu, line 128: Translation could be added to this description

frontend/src/components/modals/IntegrationOauthModal.tsx
```diff
@@ -65,38 +63,43 @@ export const OauthModal: React.FC<OauthModalProps> = ({
...
+            roadmapId,
+          );
+          if (success) {
+            closeModal(true);
+            return;
+          }
           setErrorMessage(
             `Unable to swap ${titleCase(
               name,
             )} OAuth token. Please contact an administrator if the problem persists.`,
```
IidaKainu, line 95: Translations to the error messages in IntegrationOauthModal could be added.
Here's how to add translations with variables:
https://www.i18next.com/translation-function/interpolation#working-with-data-models

frontend/src/components/modals/ImportTasksModal.tsx
```diff
@@ -109,105 +133,131 @@ export const ImportTasksModal: React.FC<ModalProps & { name: string }> = ({
...
-          )}
+              : null
+          }
+          options={boards.map((board) => ({
+            value: board.id,
+            label: board.name,
+          }))}
+        />
+        <label htmlFor="labels" style={{ marginTop: '1em' }}>
+          Select labels to import:
```
IidaKainu, line 194: Translation is missing here

frontend/src/components/modals/ImportTasksModal.tsx
```diff
@@ -109,105 +133,131 @@ export const ImportTasksModal: React.FC<ModalProps & { name: string }> = ({
...
-          <ModalFooterButtonDiv>
+  return (
+    <Form onSubmit={handleSubmit}>
+      <ModalHeader closeModal={closeModal}>
+        <h3>
+          <Trans i18nKey="Import tasks from" /> {titleCase(name)}
+        </h3>
+      </ModalHeader>
+      <ModalContent>
+        <label htmlFor="board">Select {titleCase(name)} board:</label>
```
IidaKainu, line 235: here as well

IidaKainu (approved):
The user flow is works well now! Nice modification to the modals 🔥

> I don't think this would be as common use pattern, and maybe even surprising

Now that you questioned it I think you are correct on this one.

I found a few missing translations, but otherwise this looks great!

IidaKainu (approved):

PR #102, opened 2021-07-09
Fix occasional customer corruption on update

PatricKangasmaki:
Fix an issue where occasionally customer could get corrupted in the controller, causing the updating customers to stop working

### DIFF
--- a/server/src/api/customer/customer.controller.ts
+++ b/server/src/api/customer/customer.controller.ts
@@ -65,8 +65,12 @@ export const patchCustomer: RouteHandlerFnc = async (ctx, _) => {
       .withGraphFetched('representatives');
     if (!customer) return false;
 
-    const ok = await customer.$query(trx).patch({ name, email, weight, color });
-    if (!representatives) return ok && customer;
+    if (!representatives) 
+      return await customer
+        .$query(trx)
+        .patchAndFetch({ name, email, weight, color })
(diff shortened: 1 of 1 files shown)


juhoeerola (changes requested):
As mentioned in the other comments, the logic works for the current frontend, but allowing partial patching would be desirable.

The previous logic was:
- find the correct customer
- patch the properties excluding representatives
- if no representatives
  - return patched customer, or false if nothing was patched
- patch representatives
- return patched customer

Ps. it actually seems to be the typescript update... ~I'm pretty sure the change in behaviour was introduced in the knex/objection.js update~

PatricKangasmaki:
Would this solution work better?

New logic:

- Find correct customer -> return false if not found
- If no representatives -> return updated customer (If nothing was patched -> return false)
- Patch representatives
- If none of the customer values (name, email, color, weight) were updated, return customer with new fetched representatives
- Patch customer and fetch updated representatives

Now you can update values only, representatives only or both at the same time.

juhoeerola (approved):

juhoeerola:
Seems good!

Could you add  some tests for the customer routes, to ensure it keeps working

PatricKangasmaki:
> Could you add some tests for the customer routes, to ensure it keeps working

That sounds like a good idea, I'll start working on the tests and make a new pr when they're ready 👍

PR #104, opened 2021-07-12
Fix a task table customer-dot bug

PatricKangasmaki:
Bug description: _If user who represents many customers at the same time gave rating to one of the customers, all customer dots in 'waiting for ratings' -column disappeared as if they had been rated already._

Waiting for ratings now takes "rating.forCustomer" into account as well instead of just "rating.createdByUser"

### DIFF
--- a/frontend/src/components/TableUnratedTaskRow.tsx
+++ b/frontend/src/components/TableUnratedTaskRow.tsx
@@ -72,12 +72,26 @@ export const TableUnratedTaskRow: React.FC<TableTaskRowProps> = ({ task }) => {
   */
   useEffect(() => {
     if (type === RoleType.Admin) {
-      const ratingIds = task.ratings.map((rating) => rating.createdByUser);
+      const ratings = task.ratings.map((rating) => ({
+        createdByUser: rating.createdByUser,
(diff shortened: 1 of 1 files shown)


frontend/src/components/TableUnratedTaskRow.tsx
```diff
@@ -72,12 +72,28 @@ export const TableUnratedTaskRow: React.FC<TableTaskRowProps> = ({ task }) => {
   */
   useEffect(() => {
     if (type === RoleType.Admin) {
-      const ratingIds = task.ratings.map((rating) => rating.createdByUser);
+      const ratings = task.ratings.map((rating) => {
+        return {
+          createdByUser: rating.createdByUser,
+          forCustomer: rating.forCustomer,
+        };
```
IidaKainu, line 79: This can be written without the return statement

```
const ratings = task.ratings.map((rating) => ({
  createdByUser: rating.createdByUser,
  forCustomer: rating.forCustomer,
}));
```

IidaKainu (approved):
The customer-dots now function as they are supposed to - at least in the scenarios I tried 😄

PR #105, opened 2021-07-12
Extend initial test data

IidaKainu:
Summarized the changes in the commit message.
Do comment if I missed any useful test data!

### DIFF
--- a/server/src/seeds/testdata.ts
+++ b/server/src/seeds/testdata.ts
@@ -2,6 +2,7 @@ import * as Knex from 'knex';
 import { Model } from 'objection';
 import Roadmap from '../api/roadmaps/roadmaps.model';
 import User from '../api/users/users.model';
+import Task from '../api/tasks/tasks.model';
 import { Role } from '../api/roles/roles.model';
 import Version from '../api/versions/versions.model';
 import {
@@ -9,6 +10,11 @@ import {
   TaskRatingDimension,
 } from '../../../shared/types/customTypes';
 import Customer from '../api/customer/customer.model';
+import {
+  getUser,
+  randomTaskratingValue,
+  insertRoles,
+} from '../utils/testdataUtils';
 
 export async function seed(knex: Knex): Promise<any> {
   Model.knex(knex);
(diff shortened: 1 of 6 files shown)


PatricKangasmaki:
> Do comment if I missed any useful test data!

Maybe we could use one or two extra test roadmaps if it's not too big hassle to add them, can't remember the exact cases but sometimes they would've been nice. They will probably be needed as well when we start building the "your roadmaps/projects" page which displays them all on the same page 🤔 Otherwise looking good, seems like the deletion works now 👍

IidaKainu:
> Maybe we could use one or two extra test roadmaps

Added two roadmaps 👍 
Roadmap 2 contains:
- same users, but no customers
- 3 tasks (1 unrated)
- a milestone

Roadmap 3 contains:
- only AdminPerson1 and DeveloperPerson1 as users
- one customer
- no tasks or milestones

This was fruitful: found new bugs. I'll add them to trello.
The dashboard crashes when no customers are present. And I think roadmap users are not updating in frontend as all users are displayed in people page in roadmap 3.

server/src/seeds/testdata.ts
```diff
@@ -171,35 +71,302 @@ const createTestUsers = async () => {
...
+      ratings: defaultRatings,
+    },
+    {
+      name: 'Test task 2',
+      description: 'Test desc 2',
+      createdBy,
+      ratings: defaultRatings.map(({ createdBy, createdFor, dimension }) =>
+        randomTaskratingValue(createdBy, createdFor, dimension),
+      ),
+    },
```
PatricKangasmaki, line 246: Seems really good 👍 Good catches on the bugs! I think marking a few tasks completed (for example task 2 and 4) would be a nice finishing touch and then this is ready to be merged in my opinion.

```
    {
      name: 'Test task 2',
      description: 'Test desc 2',
      createdBy,
      ratings: defaultRatings.map(({ createdBy, createdFor, dimension }) =>
        randomTaskratingValue(createdBy, createdFor, dimension),
      ),
      completed: true
    },
```

PatricKangasmaki, line 246: Seems really good 👍 Good catches on the bugs! I think marking a few tasks completed (for example task 2 and 4) would be a nice finishing touch and then this is ready to be merged in my opinion.

```
    {
      name: 'Test task 2',
      description: 'Test desc 2',
      createdBy,
      ratings: defaultRatings.map(({ createdBy, createdFor, dimension }) =>
        randomTaskratingValue(createdBy, createdFor, dimension),
      ),
      completed: true
    },
```

PatricKangasmaki (approved):

PatricKangasmaki (approved):
Test-data now has all the important things (at least what I could think of) and new tests seem to work well 👍

PR #107, opened 2021-07-15
Add tooltipIcon component.

islamnahidul:
Tooltip component added with the styles recommended in figma.


frontend/src/components/TooltipIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+import Tooltip from '@material-ui/core/Tooltip';
+import classNames from 'classnames';
+import css from './TooltipIcon.module.scss';
+
+export const TooltipIcon: React.FC<{
+  title: string;
+  children?: any;
+}> = ({ title, children }) => {
+  return (
+    <Tooltip
```
juhoeerola, line 11:
```suggestion
    <StylesProvider injectFirst>
      <Tooltip
```

The custom styles didn't seem to work, using styles provider should help (from @material-ui/core/styles)

frontend/src/pages/DashboardPage.tsx
```diff
@@ -111,9 +114,14 @@ export const DashboardPage = () => {
...
-          <p className={classes(css.header)}>
-            <Trans i18nKey="Unrated tasks" />
-          </p>
+          <div className={classes(css.titleContainer)}>
+            <h2 className={classes(css.title)}>
+              <Trans i18nKey="Unrated tasks" />
+            </h2>
+            <TooltipIcon title={t('tooltipMessage')}>
+              <InfoIcon className={classes(css.tooltipIcon, css.infoIcon)} />
+            </TooltipIcon>
```
juhoeerola, line 123:
```suggestion
            <InfoTooltip title={t('tooltipMessage')} />
```

Seeing the uses of this component, maybe the info icon should be part of the component.
In this case `InfoTooltip` or something similar, might be more descriptive name for the component.

This could allow using the children for the expanded description/info

juhoeerola (changes requested):
Few fixes, and some points to consider about the use case of the component
Otherwise looks good

frontend/src/components/TooltipIcon.tsx
juhoeerola: This is no longer necessary, as the styles provider is used on the whole application

juhoeerola (changes requested):
> maybe the info icon should be part of the component

Any thoughts on that

Otherwise looks good 👍, so after the typo fix this would be ready

The commits could also be deduplicated first, now each commit (looking at the messages) is repeated twice.

frontend/src/i18/english.ts
```diff
@@ -182,5 +182,9 @@ export const english = {
       'No problem. You can always start by making a new project. With a new project, you can start filling in tasks, milestones, and related people to your project to succeed in your product development goals.',
     'First project info 2/2':
       'Another way to start using VISDOM is to get an invitation to another existing project. Contact your project’s product owner to invite you in.',
+    tooltipMessage: 'Tooltip message placeholder', // TODO: remove it when the actual tooltip messages are added.
+    workValueTitle: 'Work / Value',
+    customerStakesTitle: 'Customer stakes in milestone',
+    unratedTaskMessage: 'Watinting for ratings'
```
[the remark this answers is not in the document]
islamnahidul: Useful if other icons(shape, color and position relative to the other components in a page)needs to used in other places

frontend/src/components/InfoTooltip.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const InfoTooltip: React.FC<{
+  title: string;
+  children?: any;
+}> = ({ title, children }) => {
+  return (
+    <Tooltip
+      classes={{
+        arrow: classNames(css.tooltipArrow),
+        tooltip: classNames(css.tooltip),
```
juhoeerola, line 16: Also noticed, this should probably use `classes`
```suggestion
        arrow: classes(css.tooltipArrow),
        tooltip: classes(css.tooltip),
```

juhoeerola (approved):

PR #108, opened 2021-07-15
Modify client weighting

IidaKainu:
Added a description to client weights page and adjusted the range of the slider to [0,2].
Modified database to use float type for customer weight.

Reset and save buttons are now disabled when no changes to customer weight have been made.

Slider labels are now not displayed with the exception of value 0. I'm not sure if I recall correctly how they were supposed to work and if the current implementation is good at all.

### DIFF
--- a/frontend/src/components/TaskValueCreatedVisualization.tsx
+++ b/frontend/src/components/TaskValueCreatedVisualization.tsx
@@ -9,6 +9,7 @@ import { RootState } from '../redux/types';
 import { Version } from '../redux/versions/types';
 import { Dot } from './Dot';
 import { totalValueAndWork } from '../utils/TaskUtils';
+import { percent } from '../utils/string';
 import css from './TaskValueCreatedVisualization.module.scss';
 
 const classes = classNames.bind(css);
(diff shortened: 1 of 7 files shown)


juhoeerola:
I think the idea was, that 0 would not disable the weighting, but it would set the weight to 0, thus ignoring the customer in question.

Also, the default should probably be 1.0, as that is the baseline of "no special weighting".

The sliders could have few more notches, at least such that 1.0 is available.

The slider could have the label displayed with "auto" all the time, and the displaying the value as a percentage (0% - 200%) might be an option.

IidaKainu:
> I think the idea was, that 0 would not disable the weighting, but it would set the weight to 0, thus ignoring the customer in question.

Edited the description to be more clear about this
Modified the sliders to have more notches with the default weight of 1.0. Percentages are displayed in the label (with the label="auto" all the time) 👍

juhoeerola (changes requested):
Couple small things, otherwise looks good 👍

IidaKainu:
I would not have thought of localizing percentages, good that you brought it up 👍 I added it to the utils
Also fixed the useEffect

juhoeerola (approved):

PR #109, opened 2021-07-20
Fix GET taskratings

juhoeerola:
Only fetch ratings belonging to the task.

Also added ignore for all `node_modules` and `.env` files


IidaKainu (approved):
Looks good to me 👍

PR #110, opened 2021-07-20
Task utils & translations

juhoeerola:
Updated task utils to work nicer with the current needs.

Also updated translation keys to exclude special punctuation.
Removed few unused, and added few missing ones.


frontend/src/components/RoadmapOverview.tsx
```diff
@@ -31,20 +31,11 @@ export const RoadmapOverview = () => {
...
-        }
-      });
-    });
-
-    return count > 0 ? sum / count : 0;
-  };
+  const { value, work } = averageValueAndWork(roadmap?.tasks ?? []);
+  const numFormat = new Intl.NumberFormat(undefined, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 2,
```
IidaKainu, line 37: The numFormat could be moved to /utils/string.ts (perhaps string.ts should be renamed to format.ts?)
```
export const numFormat = (
  minimumFractionDigits?: number,
  maximumFractionDigits?: number,
) =>
  new Intl.NumberFormat(undefined, {
    minimumFractionDigits,
    maximumFractionDigits,
  });
```

frontend/src/utils/TaskUtils.ts
```diff
@@ -44,94 +44,70 @@ export enum SortingTypes {
...
+    return this.sum;
+  }
+  get avg() {
+    return this.count ? this.sum / this.count : 0;
+  }
+  add(rating: number) {
+    this.sum += rating;
+    this.count += 1;
+    return this;
+  }
```
IidaKainu, line 60: Eslint complaints about expecting a blank line between class members here. Maybe the rule could be disabled in .eslintrc.js rules:
`'lines-between-class-members': 'off',`

IidaKainu (changes requested):
The task utils look good and also more readable for me.
Found only a few things to comment on.

frontend/src/utils/TaskUtils.ts
PatricKangasmaki: In my opinion, some spacing would help with code readability. For example, one liners could be without spacing and multi-lines could be separated?

`    'lines-between-class-members': [
      'error',
      'always',
      { 'exceptAfterSingleLine': true },
    ],`
    
It would then look like


```
class RatingsSummary {
  private sum: number = 0;
  private count: number = 0;

  get total() {
    return this.sum;
  }

  get avg() {
    return this.count ? this.sum / this.count : 0;
  }

  add(rating: number) {
    this.sum += rating;
    this.count += 1;
    return this;
  }
};
```

juhoeerola: I added blank lines, I think that helps the readability

I just had eslint off locally, and unfortunately prettier won't add those lines

frontend/src/components/RoadmapOverview.tsx
juhoeerola: I did consider that, but it would end up very thin wrapper around the constructor, with more limited options.

Such, specialised wrapper can be good, as it focuses the use case further, but I'm not sure if that is the case here.

IidaKainu (approved):

PR #111, opened 2021-07-20
Customer api tests

PatricKangasmaki:
Customer deletion caused `ForeignKeyViolationError` so I had to create a new migration to fix that.
I also added some tests for customers, tell me if there's too many or too little of them 😄 Tried to add tests for the most common situations


PatricKangasmaki:
Forgot `.only` in customer tests so none of the other test suites ran, previous force-push was just a quick fix for that

IidaKainu (changes requested):
In my limited opinion the amount of tests here is fine 👍 
I think it's nice how you isolated the GET, POST, PATCH, DELETE tests to not depend on other routes (instead fetching data from db).

server/src/migrations/20210719114220_addCustomerIdCascadeOnDelete.ts
```diff
@@ -0,0 +1,25 @@
+import * as Knex from "knex";
```
juhoeerola, line 1: I'd guess you don't have prettier setup to format on save (it would change to single quotes)

It might be good to do, to avoid unnecessary formatting noise later

juhoeerola:
> Forgot `.only` in customer tests

You can also give extra options to `yarn test`, which are passed to `mocha`, e.g.

```sh
yarn test -g 'regex to match'  # run only matching tests
yarn test --help  # to see more options
```

server/tests/customer.test.ts
```diff
@@ -0,0 +1,404 @@
...
+
+      const customersBefore = await Customer.query().where('roadmapId', firstRoadmapId);
+      const res1 = await postCustomer(invalidName);
+      const res2 = await postCustomer(invalidEmail);
+      const res3 = await postCustomer(invalidColor);
+      const customersAfter = await Customer.query().where('roadmapId', firstRoadmapId);
+
+      expect(res1.status).to.equal(400);
+      expect(res2.status).to.equal(400);
+      expect(res3.status).to.equal(400);
```
juhoeerola, line 148: This kind of approach has few slight disadvantages:
- only the first failure is shown
- the failing assertion is not clear in the logs (one has to check the reported line number)

```suggestion
      const res1 = await postCustomer(invalidName);
      const res2 = await postCustomer(invalidEmail);
      const res3 = await postCustomer(invalidColor);
      const customersAfter = await Customer.query().where('roadmapId', firstRoadmapId);

      expect(res1.status, 'name').to.equal(400);
      expect(res2.status, 'email').to.equal(400);
      expect(res3.status, 'color').to.equal(400);
```

The latter can be improved by adding message to the assertion.

Although, both can be improved with parametrised test, an example of those is in
https://github.com/Vincit/VISDOM-Roadmapper/blob/1cdf7b6a11a244c511c6b4c519151600338bf45e/server/tests/users.test.ts#L134-L149

server/src/migrations/20210719114220_addCustomerIdCascadeOnDelete.ts
PatricKangasmaki: Ah, didn't know about the format on save, I set it up now, thanks!

server/src/utils/testUtils.ts
```diff
@@ -0,0 +1,44 @@
+import { loggedInAgent } from '../../tests/setuptests';
+import Roadmap from '../api/roadmaps/roadmaps.model';
+import Customer from '../api/customer/customer.model';
+import User from '../api/users/users.model';
+import { Role } from '../api/roles/roles.model';
+import { Permission, RoleType } from '../../../shared/types/customTypes';
+
+export const removePermission = async (permission: Permission) => {
+  const firstRoadmapId = (await Roadmap.query().first()).id;
+  const userId = (
```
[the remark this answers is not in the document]
PatricKangasmaki: I moved the getTestRatingData to utils now and changed all the tests to use removePermission!

server/tests/customer.test.ts
PatricKangasmaki: I changed all the res1, res2 etc. implementations to use parametrised tests, works way more logically now 👍

IidaKainu (approved):

juhoeerola (changes requested):
The testUtils file could be moved to the tests directory, as those are no use for the rest of the code

Apart from the one concern, this looks great 👍

server/src/utils/testUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+    await User.query().where({ username: 'AdminPerson1' }).first()
+  ).id;
+
+  await Role.query().patchAndFetchById([userId, firstRoadmapId], {
+    type: RoleType.Admin & ~permission,
+  });
```
juhoeerola, line 17: Just noticed...

This should really use the previous permission to remove from, not the Admin permission.

Now this will not work as intended when called multiple times to remove multiple permissions.

juhoeerola (changes requested):

server/src/utils/testUtils.ts
PatricKangasmaki: I completely missed that. Would this change make sense? Now the function would take a RoleType as a parameter and the function caller could remove as many permissions from the original role as they want. Caller could also remove permissions from any role (dev, business etc.), not just admin.

Function:
```
export const removePermission = async (restrictedRole: RoleType) => {
  const firstRoadmapId = (await Roadmap.query().first()).id;
  const userId = (
    await User.query().where({ username: 'AdminPerson1' }).first()
  ).id;

  await Role.query().patchAndFetchById([userId, firstRoadmapId], {
    type: restrictedRole,
  });
};
```

Usage:
```
await removePermission(
   RoleType.Admin & ~Permission.RoadmapReadUsers & ~Permission.TaskRead,
);
```

From what I've tested, it seems to work well. I just figured it would be better to get it peer-reviewed before going around every file and change them to use this. I made a temporary fixup commit where the function is called `removePermission2` and it's used in `tests/tasks.test.ts -> GET -> Should not get tasks with incorrect permissions` if you want to try the functionality.

server/src/utils/testUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+export const updateCustomer = async (customer: Customer, newData: any) => {
+  const firstRoadmapId = (await Roadmap.query().first()).id;
+  return await loggedInAgent
+    .patch(`/roadmaps/${firstRoadmapId}/customers/${customer.id}`)
```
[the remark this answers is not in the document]
PatricKangasmaki: I added a `Roadmap.query().where('id', customer.roadmapId)` query there so the to-be-updated customer always belongs to the roadmap

server/tests/tasks.test.ts
```diff
@@ -65,19 +64,16 @@ describe('Test /roadmaps/:roadmapId/tasks/ api', function () {
       expect(added.createdByUser).to.equal(userId);
     });
     it('Should not add new task with incorrect permissions', async function () {
+      await removePermission(Permission.RoadmapReadUsers);
+      await removePermission(Permission.TaskCreate);
```
juhoeerola, line 68: These should be at the same position as the previous permission patching.

Now the `before` result is also fetched with reduced permissions.

This should probably be done in each of these cases.


Maybe this could be done easier with *kind of* context manager

Instead of
```ts
await removePermission(permission);
const res = /* whatever query */
/* remember to restore previous permission */
```

```ts
const res = await withoutPermission(permission, () => {
  return /* whatever query */
});
```

The implementation would be relatively simple:

```ts
export const withoutPermission = async <T>(
  permission: Permission,
  work: () => Promise<T>,
) => {
  const firstRoadmapId = (await Roadmap.query().first()).id;
  const userId = (
    await User.query().where({ username: 'AdminPerson1' }).first()
  ).id;

  const role = await Role.query().findById([userId, firstRoadmapId]);
  const original = role.type;
  await role.$query().patch({ type: role.type & ~permission });
  try {
    return await work();
  } finally {
    await Role.query().patchAndFetchById([userId, firstRoadmapId], {
      type: original,
    });
  }
};
```

This would also work for removing multiple permissions by:
- combining permissions to remove `a | b | c`
- or nesting `withoutPermission(a, () => withoutPermission(b, () => /* ... */))` (arguably not so nice)

juhoeerola (changes requested):
Noticed some errors in the ordering of permission changes

And added few suggestion for the `removePermission` functionality

server/src/utils/testUtils.ts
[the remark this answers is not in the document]
juhoeerola: This should fix the issue 👍 
Did you check if any of the other had similar assumptions?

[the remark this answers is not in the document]
juhoeerola: Looking at the other helper functions, those make similar assumptions

juhoeerola:
I added a possible implementation, could you take a look if that would work

PatricKangasmaki:
> I added a possible implementation, could you take a look if that would work

This implementation seems to work just as wanted, nice work 👍 I agree that removing permissions like `a | b | c` looks way cleaner than nesting. I noticed that you fixed helper functions as well, to me everything looks correct now.

juhoeerola (approved):

PR #112, opened 2021-07-21
Fix rating customers bug

PatricKangasmaki:
Rate modal in task table would show customers from every roadmap if you represented them instead of showing the ones that belong to the chosen roadmap.


IidaKainu (approved):
Look good and works correctly 👍

PR #113, opened 2021-07-21
Fix button styles

IidaKainu:
Buttons now have focus effects (incl. stepIndicator, dropdown).
Removed the extra focusable element on Checkboxes and radio buttons.

It was hard to notice the small outlined button was focused on, so I added a background color to focus/hover. Not sure if it's the most fitting solution, feedback is welcome

Keyboard navigation seemed ok to me.

### DIFF
--- a/frontend/src/components/RoadmapSelectorWidget.module.scss
+++ b/frontend/src/components/RoadmapSelectorWidget.module.scss
@@ -15,15 +15,16 @@
   background-color: transparent;
   justify-content: space-between;
   padding: 0px 16px;
-}
 
-.dropButton:focus {
-  outline: none;
-}
-
-.dropButton:disabled {
-  color: $COLOR_BLACK40;
-  cursor: not-allowed;
+  &:hover,
+  &:focus {
+    outline: none;
+    background: white;
+  }
+  &:disabled {
+    color: $COLOR_BLACK40;
+    cursor: not-allowed;
+  }
 }
 
 .dropMenu {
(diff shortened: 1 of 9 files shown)


PatricKangasmaki (changes requested):
These hovers are a good idea! 

I just think the colors need to be adjusted a bit in the outlined buttons, currently the border gets darker and background lighter, feels a bit off to me personally. I was playing around with the colors and in my not-designer-opinion, you could keep the border same color as non-hovered and change the background to either lighter or darker. Tested darker as well because full white looked a bit too bright (Might be because I'm used to using dark mode everywhere though 😄)

Here's some tests that I did, buttons on top are hovered and bottom non-hovered. Same thing applies to the RoadmapSelector as well.
suggestion

Just a thought: would these milestone hovers need some reworking as well? Currently the top blends with the background when they're hovered. Maybe changing the color a bit would help so it's not the same as background and you could tell where the box ends.

hover1

IidaKainu:
> currently the border gets darker and background lighter, feels a bit off to me personally

That felt weird to me too, but I could not point out why. Thanks! I went with the middle one of your examples. (the darker grey background somehow gave disabled-vibes)

For milestone hovers, I added border and a slight scaling effect and removed the background color change.
I think it might be fitting since figma has colored borders for different milestone states (of course the hover color needs to be different for those)

PatricKangasmaki (approved):
Looking good 👍 I also tested the keyboard navigation and it seemed to work well

PR #114, opened 2021-07-21
Fix dashboard crash when there's no customers

PatricKangasmaki:
Dashboard would break because functions `calcTaskWeightedValueSum `and `unratedProductOwnerTasks `won't work when customers are empty. I added customer checks in the functions (works now) but I was wondering if this solution is the correct way? Or should the customer empty/undefined/null checks be in the components where you call these functions?

### DIFF
--- a/frontend/src/components/PlannerChart.tsx
+++ b/frontend/src/components/PlannerChart.tsx
@@ -59,8 +59,8 @@ export const PlannerChart: React.FC<{
       name: DataKeys.OptimalRoadmap,
       tasks: [...currentRoadmap.tasks].sort(
         (a, b) =>
-          calcWeightedTaskPriority(b, customers!, currentRoadmap) -
-          calcWeightedTaskPriority(a, customers!, currentRoadmap),
+          calcWeightedTaskPriority(b, customers ?? [], currentRoadmap) -
+          calcWeightedTaskPriority(a, customers ?? [], currentRoadmap),
       ),
     });
   }
--- a/frontend/src/components/TaskTable.tsx
+++ b/frontend/src/components/TaskTable.tsx
@@ -161,10 +161,15 @@ export const TaskTableUnrated: React.FC<{
   useEffect(() => {
     const type = getType(userInfo?.roles, currentRoadmap?.id);
     if (type === RoleType.Admin) {
-      if (userInfo && currentRoadmap && allUsers)
+      if (userInfo && currentRoadmap && allUsers) {
         setTaskList(
-          unratedProductOwnerTasks(tasks, allUsers, currentRoadmap.customers),
+          unratedProductOwnerTasks(
+            tasks,
+            allUsers,
+            currentRoadmap.customers ?? [], // FIXME: incorrect type, customers may be undefined
+          ),
         );
+      }
     }
     if (type === RoleType.Developer || type === RoleType.Business) {
       setTaskList(tasks.filter((task) => taskAwaitsRatings(task, userInfo)));
@@ -217,7 +222,7 @@ export const TaskTableRated: React.FC<{
         const unratedTasks = unratedProductOwnerTasks(
           tasks,
           allUsers,
-          currentRoadmap.customers,
+          currentRoadmap.customers ?? [], // FIXME: incorrect type, customers may be undefined
         );
         setTaskList(tasks.filter((task) => !unratedTasks?.includes(task)));
       }


frontend/src/utils/TaskUtils.ts
```diff
@@ -245,6 +245,8 @@ export const calcTaskWeightedValueSum = (
   allCustomers: Customer[],
   roadmap: Roadmap,
 ) => {
+  if (!allCustomers) return;
```
juhoeerola, line 248: The type is `Customer[]`, which will never be falsy, so the correct fix would either be:
- also change the parameter type
- check the parameter at call site

This would also change the return type from `number` to `number | undefined`, which would require additional handling at the call site.

I would suggest that the "correct" fix would be checking the parameters in the caller (e.g. `customers ?? []`, if the caller doesn't care about undefined customers)

juhoeerola (changes requested):
I think this is part of larger problem

There are some incorrect types for the objects returned from the backend, where some fields could really also be `undefined`, but the type doesn't account for that.

Also, in a bit too many places the frontend just does non-null type assertions, where those would not really be justified.

I think the "correct" fix would be checking the possible nulls more carefully, and fixing the types to reflect the reality.

Adding the checks incrementally to components where crashes are noticed would be reasonable approach for this. When doing such fixes, the offending type should also be noted down, unless it is fixed at the same time. (otherwise it's harder to track these down later)

PatricKangasmaki:
I moved the fixes to the components where the functions are called now and left a comment about what caused the crash if there's a larger fix in the future

frontend/src/components/PlannerChart.tsx
```diff
@@ -55,12 +55,13 @@ export const PlannerChart: React.FC<{
...
 
+    // const customers: Customer[] | undefined   <- breaks taskUtils/calcWeightedTaskPriority if undefined
     graphTaskLists.unshift({
       name: DataKeys.OptimalRoadmap,
       tasks: [...currentRoadmap.tasks].sort(
         (a, b) =>
-          calcWeightedTaskPriority(b, customers!, currentRoadmap) -
-          calcWeightedTaskPriority(a, customers!, currentRoadmap),
+          calcWeightedTaskPriority(b, customers ?? []!, currentRoadmap) -
+          calcWeightedTaskPriority(a, customers ?? []!, currentRoadmap),
```
juhoeerola, line 64: Here the behaviour was consistent with the existing types, e.i. `customers` was typed as `Customer[] | undefined`
The problem was the incorrect type assertion, so I'd say the comment is not needed here

Also, the assertion is now unnecessary
```suggestion
          calcWeightedTaskPriority(b, customers ?? [], currentRoadmap) -
          calcWeightedTaskPriority(a, customers ?? [], currentRoadmap),
```

juhoeerola, line 64: Here the behaviour was consistent with the existing types, e.i. `customers` was typed as `Customer[] | undefined`
The problem was the incorrect type assertion, so I'd say the comment is not needed here

Also, the assertion is now unnecessary
```suggestion
          calcWeightedTaskPriority(b, customers ?? [], currentRoadmap) -
          calcWeightedTaskPriority(a, customers ?? [], currentRoadmap),
```

frontend/src/components/TaskTable.tsx
```diff
@@ -161,10 +161,16 @@ export const TaskTableUnrated: React.FC<{
...
     if (type === RoleType.Admin) {
-      if (userInfo && currentRoadmap && allUsers)
+      if (userInfo && currentRoadmap && allUsers && tasks) {
+        // currentRoadmap.customers breaks taskUtils/unratedProductOwnerTasks if it's undefined
         setTaskList(
-          unratedProductOwnerTasks(tasks, allUsers, currentRoadmap.customers),
+          unratedProductOwnerTasks(
+            tasks,
+            allUsers,
+            currentRoadmap.customers ?? [],
```
juhoeerola, line 170: This is actual type problem 👍 

The comment could be attached to this line, and be more specific to the problem, e.g.
```suggestion
            currentRoadmap.customers ?? [], // FIXME: incorrect type, customers may be undefined
```

Similar comment could maybe be added to the corresponding type definition (in redux/roadmaps/types), until it gets fixed

frontend/src/components/TaskTable.tsx
```diff
@@ -161,10 +161,16 @@ export const TaskTableUnrated: React.FC<{
   useEffect(() => {
     const type = getType(userInfo?.roles, currentRoadmap?.id);
     if (type === RoleType.Admin) {
-      if (userInfo && currentRoadmap && allUsers)
+      if (userInfo && currentRoadmap && allUsers && tasks) {
```
juhoeerola, line 164: The `tasks` is `Task[]`, so the check should be unnecessary, and possible null checks should belong to the user of this component

```suggestion
      if (userInfo && currentRoadmap && allUsers) {
```

juhoeerola (changes requested):
Couple suggestions

I also added a card to trello, for tracking these incorrect types, so we can update that when noticing new ones

PatricKangasmaki:
Added all your suggestions 👍

juhoeerola (approved):

PR #115, opened 2021-07-21
Add scaling to RoadmapGraphPage's total values

IidaKainu:
Multiply a client's ratings by their weight (or planner weight) before calculating the taskrating total value.

The scaled total value is displayed only in the roadmap graph versions. Should it be displayed in milestones editor too?
 

Height is calculated logarithmically: the full height of 350 is reached when a version's total value is a bit over 200.
`const h = Math.min(350, Math.max(100, 45 * Math.log2(ver.value)));`

### DIFF
--- a/frontend/src/utils/TaskUtils.ts
+++ b/frontend/src/utils/TaskUtils.ts
@@ -220,43 +220,39 @@ export const dragDropBetweenLists = (
   };
 };
 
-export const calcTaskWeightedValueSum = (
+const ratingValue = (allCustomers: Customer[], roadmap: Roadmap) => (
+  rating: Taskrating,
+) => {
+  const ratingCreator = allCustomers.find(
+    ({ id }) => id === rating.forCustomer,
+  );
+  const creatorWeight = ratingCreator
+    ? customerWeight(ratingCreator, roadmap.plannerCustomerWeights)
+    : 0;
+
+  return rating.value * creatorWeight;
+};
+
+const taskWeightedValueSummary = (
   task: Task,
   allCustomers: Customer[],
   roadmap: Roadmap,
-) => {
-  const customerValuesSum = allCustomers.reduce(
-    (total, customer) => total + customerWeight(customer),
-    0,
-  );
-
-  const ratingValue = (rating: Taskrating) => {
-    const ratingCreator = allCustomers.find(
-      ({ id }) => id === rating.forCustomer,
-    );
-
-    const ratingCreatorValue = ratingCreator
-      ? customerWeight(ratingCreator, roadmap.plannerCustomerWeights)
-      : 0;
-
-    let creatorPlannerWeight = roadmap.plannerCustomerWeights?.find(
-      ({ customerId }) => customerId === rating.forCustomer,
-    )?.weight;
-    if (creatorPlannerWeight === undefined) creatorPlannerWeight = 1;
-
-    let creatorValueWeight = 0;
-    if (ratingCreator === undefined) creatorValueWeight = 1;
-
-    if (ratingCreatorValue > 0 && customerValuesSum > 0) {
-      creatorValueWeight = ratingCreatorValue / customerValuesSum;
-    }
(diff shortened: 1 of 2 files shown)


frontend/src/utils/TaskUtils.ts
```diff
@@ -220,51 +220,64 @@ export const dragDropBetweenLists = (
...
+  allCustomers: Customer[],
+  roadmap: Roadmap,
+) =>
+  task.ratings
+    .filter(({ dimension }) => dimension === TaskRatingDimension.BusinessValue)
+    .map(ratingValue(allCustomers, roadmap))
+    .reduce((acc, num) => {
+      const previous = acc ?? new RatingsSummary();
+      return previous.add(num);
+    }, result);
```
juhoeerola, line 248:
```suggestion
    .reduce((acc, num) => acc.add(num), result);
```

either there is no need for the conditional
or the parameter type is incorrect

frontend/src/utils/TaskUtils.ts
```diff
@@ -220,51 +220,64 @@ export const dragDropBetweenLists = (
...
-    let creatorPlannerWeight = roadmap.plannerCustomerWeights?.find(
-      ({ customerId }) => customerId === rating.forCustomer,
-    )?.weight;
-    if (creatorPlannerWeight === undefined) creatorPlannerWeight = 1;
+  return rating.value * creatorWeight;
+};
 
-    let creatorValueWeight = 0;
-    if (ratingCreator === undefined) creatorValueWeight = 1;
+export const taskWeightedValueSummaryInto = (
```
juhoeerola, line 236: Is there a need for both this and `taskWeightedValueSummary`?

Also both are only used internally, so maybe these could be not exported, unless there is use case in sight

IidaKainu: I guess it was rather useless to have taskWeightedValueSummary to only provide the RatingsSummary. Removed!
Removed the export from the other function as well

frontend/src/utils/TaskUtils.ts
IidaKainu: Thank you for pointing this out: it's now corrected

frontend/src/utils/TaskUtils.ts
```diff
@@ -220,51 +220,61 @@ export const dragDropBetweenLists = (
...
-      creatorValueWeight = ratingCreatorValue / customerValuesSum;
-    }
-    return rating.value * creatorValueWeight * creatorPlannerWeight;
-  };
+  return rating.value * creatorWeight;
+};
 
-  return task.ratings
+const taskWeightedValueSummary = (
+  result: RatingsSummary,
```
juhoeerola, line 237: The `result` parameter might be simpler to remove, and just create `new RatingsSummary()` in this function.

Or did you have some use case in mind, where the result would be needed as parameter?

IidaKainu: No, I can't think of a case when it would be needed. 
That's really only the side-effect of me not comprehending ratingsSummaryByDimensionInto and trying to mimic its usage. I now see the solution is actually simple... (and understand what exactly is going on in ratingsSummaryByDimensionInto)

juhoeerola:
The current scaling looks good 👍 

> Should it [the scaled value] be displayed in milestones editor too?

Should the customer stakes visualisation also be scaled? It feels a bit odd now, as the other visualisation on that page changes, but this stays constant. (the same on dashboard page with the heat map and graph)

It might be good to consistently use scaled values everywhere, so all the visualisations and numbers stay consistent.
In that case there could be some convenient way to toggle the scaling off (e.i. to 1.0 for all)

juhoeerola (approved):

IidaKainu:
> Should the customer stakes visualisation also be scaled? 

I agree: I made a card about it so it can be done separately. I am merging this pr since it's approved.

> It might be good to consistently use scaled values everywhere

This would be also nice, we just need to think about how to make the toggle convenient.

PR #117, opened 2021-07-26
Project overview page

IidaKainu:
Added the project overview page layout. 

Modifying the redux selectors turned out to be a bigger task, so the project summaries are missing some data. 
I am planning on creating a separate for the redux modifications. However, I can draft this pr if the page is too incomplete for merging!


frontend/src/pages/ProjectOverviewPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <div
+            className={classes(css.addRoadmap)}
+            role="button"
+            onClick={addRoadmapClicked}
+            onKeyPress={addRoadmapClicked}
+            tabIndex={0}
+          >
+            <ControlPointSharpIcon />
+            <Trans i18nKey="Add project" />
+          </div>
```
juhoeerola, line 73: This style of button could be made into a separate component, as it is used in multiple places in the design

At the moment I have this in the #116

frontend/src/components/MetricsSummary.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Trans } from 'react-i18next';
+import css from './MetricsSummary.module.scss';
+
+const classes = classNames.bind(css);
+
+export const MetricsSummary: React.FC<{
```
juhoeerola, line 8: 👍

frontend/src/components/MetricsSummary.tsx
```diff
@@ -0,0 +1,21 @@
...
+  label: string;
+  value: string | number;
+  children?: any;
+}> = ({ label, value, children }) => (
+  <div className={classes(css.dataEntryWrapper)}>
+    <Trans i18nKey={label} />
```
juhoeerola, line 14: I'm not quite sure if the label should be translated by the user of this component.
That might lead to less possible confusion

frontend/src/shared.scss
```diff
@@ -31,6 +31,14 @@
   box-shadow: 0px 0px 0px 3px rgba(255, 48, 97, 0.25);
 }
 
+.visdomLogo {
+  cursor: pointer;
+  &:hover,
+  &.selected {
+    filter: brightness(0.8);
+  }
```
juhoeerola, line 39: I'm not quite sure if the filter should stay when selected
To my eye, the dim logo doesn't look quite as good

juhoeerola (commented):
The three dot menu could be added
And the 💚 , which indicates the default project, I guess...

Also, the client list is empty until the project is opened once. Is this a limitation of the current data model?

frontend/src/shared.scss
PatricKangasmaki: In my opinion the dimmed logo is good while it's hovered but I agree that it looks a bit off when it stays dimmed when selected. But I think there should be some other indication that you're on this page, maybe something similar to the other buttons on the navbar?

Also, does the user know that the logo in the corner is clickable button to the projects-page or should there be another way to get there/some indication (other than the hover) that you can indeed click it?

juhoeerola: >some other indication that you're on this page

The lack of indication on the other sidebar items could be taken as indication

(It might even make sense to not show the sidebar on the overview page, as all those items are roadmap specific.
Then the overview page would be the home page for the whole application, and dashboard page would be the home page for a roadmap.)

>does the user know that the logo in the corner is clickable button to the projects-page

I think the logo at the top left is quite idiomatic as link to the "home" page. This might be more obvious if the user would be directed on that page on login.

Maybe the "home" page resolution after login could be taking the first matching:
1. redirect in url
2. default roadmap
3. the overview page

PatricKangasmaki: > It might even make sense to not show the sidebar on the overview page, as all those items are roadmap specific.

That would make sense, it would separate homepage from roadmap-specific pages nicely. I think this is worth trying, it could look and feel really good. I was playing around a bit with this idea, thoughts?


idea

You'd probably have to change the functionality a bit though in this case since you really can't have a disabled open -button on your selected roadmap if the only way to get to roadmaps is the open -button. Maybe going to the homepage would set your selected roadmap to none.

> 1. redirect in url 2.default roadmap 3. the overview page

That sounds like a good idea

frontend/src/components/RoadmapOverview.tsx
```diff
@@ -38,35 +38,15 @@ export const RoadmapOverview = () => {
...
-      </div>
+    <div className={classes(css.data)}>
+      <MetricsSummary label="Tasks" value={roadmapTasksCount()} />
+      <MetricsSummary label="Milestones" value={roadmapMilestonesCount()} />
+      <MetricsSummary label="Avg Value" value={numFormat.format(value)}>
+        <MonetizationOnIcon />
+      </MetricsSummary>
+      <MetricsSummary label="Avg Work" value={numFormat.format(work)}>
+        <WorkIcon />
+      </MetricsSummary>
```
PatricKangasmaki, line 49: I was thinking if this kind of approach would be cleaner since you wouldn't have to repeat the same component X-times and adding a new metric would be easier. Or would the metrics-array make it look messier, what do you think?

```suggestion
  const metrics = [
    {
      label: 'Tasks',
      metricsValue: roadmapTasksCount(),
    },
    {
      label: 'Milestones',
      metricsValue: roadmapMilestonesCount(),
    },
    {
      label: 'Avg Value',
      metricsValue: numFormat.format(value),
      children: <MonetizationOnIcon />,
    },
    {
      label: 'Avg Work',
      metricsValue: numFormat.format(work),
      children: <WorkIcon />,
    },
  ];

  return (
    <div className={classes(css.data)}>
      {metrics.map(({ label, metricsValue, children }) => (
        <MetricsSummary key={label} label={label} value={metricsValue}>
          {children}
        </MetricsSummary>
      ))}
    </div>
  );
```

frontend/src/components/MetricsSummary.tsx
IidaKainu: labels are now translated by the user, not the component itself 👍

frontend/src/shared.scss
IidaKainu: For now I removed the selected filter effect on the Visdom logo.
The overview page works better without the sidebar 👍 I however think the top bar should stay the same, perhaps without the project selector dropdown.

IidaKainu:
Added the three dot menu
- the "delete roadmap" link inside it will look and behave as its supposed to once href is provided

Added default project 💚 although it's not used yet. I think we should add a default project boolean value to roadmap model.

> the client list is empty until the project is opened once. Is this a limitation of the current data model?

Yes. I got that fixed, but I will submit it in a different pr to keep this simple.

Added all of the requested fixes.

About the "AddButton" component:
- figma had differing default colors for it (grey in overview page, green otherwise). I went with green as it is more distinct from the disabled color.
- It's a mix between Juho's and my implementations
- Currently the button itself is quite small, but it will stretch to fit the container it's inside of. This does leave some responsibility for the user to remember to use it inside a styled div.
- Does anyone know how to override the material ui styles without adding !importart to scss?

PatricKangasmaki:
> Does anyone know how to override the material ui styles without adding !importart to scss?

I used `<StylesProvider injectFirst>` to override material ui tooltip styling with our figma design in task tables, seemed to work fine.
https://material-ui.com/styles/advanced/#injectfirst

PatricKangasmaki (approved):
This looks good to me now 👍 Wait for Juho's approval too before merging though 😄

juhoeerola (approved):
Looks great 👍 

Noticed a couple more things, but after checking those, this should be good to merge

Also, before merging, some of the commits could maybe be squashed a bit

frontend/src/components/ProjectSummary.tsx
```diff
@@ -0,0 +1,134 @@
...
+          tabIndex={0}
+          type="button"
+        >
+          <Trans i18nKey="Set as default project" />
+        </button>
+        {/* TODO: href to roadmap deletion modal */}
+        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+        <a className={classes(css.green)}>
+          <Trans i18nKey="Delete project..." />
+        </a>
```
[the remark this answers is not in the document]
IidaKainu: The link will have pointer cursor once the href is provided 👍 This is only a problem for incomplete links like this

PR #119, opened 2021-07-28
Modify frontend customers

IidaKainu:
Customers can be selected by the id of a roadmap in addition to the
selected/current roadmap id in redux.
Customers from other roadmaps can be fetched in frontend api.


juhoeerola (approved):
👍 seems to work fine

PR #120, opened 2021-07-28
Add StylesProvider to App

IidaKainu:
StylesProvider ables overriding of Material-UI's styles. It is added to App root so styles can be overridden easier.


juhoeerola (approved):
👍

PR #121, opened 2021-07-28
Add notification modal for missing ratings

PatricKangasmaki:
Notifications modal for users that haven't given ratings when they should (it parses out logged in user, I figured that you don't want to send email to yourself). 
Currently it sends the api request to missing endpoint so it should give 404. 
Payload contains: List of users that were checked and the correct task.

### DIFF
--- a/frontend/src/api/api.tsx
+++ b/frontend/src/api/api.tsx
@@ -306,6 +306,18 @@ const patchIntegrationConfiguration = async (
   return response.data as IntegrationConfiguration;
 };
 
+const sendNotification = async (
+  users: number[],
+  task: Task,
+  message: string,
+) => {
+  const response = await axios.post(
+    `roadmaps/${task.roadmapId}/tasks/${task.id}/notify`,
+    { users, message },
+  );
+  return response.status === 200;
+};
+
 export const api = {
   getRoadmaps,
   addRoadmap,
@@ -342,4 +354,5 @@ export const api = {
   swapIntegrationOAuthToken,
   addIntegrationConfiguration,
   patchIntegrationConfiguration,
+  sendNotification,
 };
--- a/frontend/src/components/modals/NotifyUsersModal.tsx
+++ b/frontend/src/components/modals/NotifyUsersModal.tsx
@@ -1,0 +1,230 @@
+import React, { useState, useEffect } from 'react';
+import { Alert, Form } from 'react-bootstrap';
+import { Trans } from 'react-i18next';
+import { useDispatch, useSelector, shallowEqual } from 'react-redux';
+import classNames from 'classnames';
+import { roadmapsActions } from '../../redux/roadmaps';
+import { StoreDispatchType } from '../../redux';
+import {
+  allCustomersSelector,
+  roadmapUsersSelector,
+  taskSelector,
+} from '../../redux/roadmaps/selectors';
+import { LoadingSpinner } from '../LoadingSpinner';
+import { ModalContent } from './modalparts/ModalContent';
+import { ModalFooter } from './modalparts/ModalFooter';
+import { ModalFooterButtonDiv } from './modalparts/ModalFooterButtonDiv';
+import { ModalHeader } from './modalparts/ModalHeader';
+import { Checkbox } from '../forms/Checkbox';
+import { RootState } from '../../redux/types';
+import { UserInfo } from '../../redux/user/types';
+import { userInfoSelector } from '../../redux/user/selectors';
+import { Customer, RoadmapUser } from '../../redux/roadmaps/types';
+import { RoleType } from '../../../../shared/types/customTypes';
+import css from './NotifyUsersModal.module.scss';
+import { getType } from '../../utils/UserUtils';
+import { Modal, ModalTypes } from './types';
+import {
+  findMissingCustomers,
+  findMissingDevelopers,
+} from '../../utils/TaskUtils';
+import { TextArea } from '../forms/FormField';
+
+const classes = classNames.bind(css);
+
+interface CheckableUser extends RoadmapUser {
+  checked: boolean;
+}
+
+export const NotifyUsersModal: Modal<ModalTypes.NOTIFY_USERS_MODAL> = ({
+  closeModal,
+  taskId,
+}) => {
+  const dispatch = useDispatch<StoreDispatchType>();
+  const task = useSelector(taskSelector(taskId))!;
+  const [isLoading, setIsLoading] = useState(false);
+  const userInfo = useSelector<RootState, UserInfo | undefined>(
+    userInfoSelector,
+    shallowEqual,
+  );
+  const customers = useSelector<RootState, Customer[] | undefined>(
+    allCustomersSelector(),
+    shallowEqual,
+  );
+  const allUsers = useSelector<RootState, RoadmapUser[] | undefined>(
+    roadmapUsersSelector(),
+    shallowEqual,
+  );
+  const [missingUsers, setMissingUsers] = useState<CheckableUser[] | undefined>(
+    [],
+  );
+  const [previousUsers, setPreviousUsers] = useState<
+    CheckableUser[] | undefined
+  >([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [message, setMessage] = useState('');
+  const [allChecked, setAllChecked] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (
+      getType(userInfo?.roles, task.roadmapId) === RoleType.Admin &&
+      customers &&
+      allUsers
+    ) {
+      const missingCustomers = findMissingCustomers(task.ratings, customers);
+      const missingRepresentatives = new Map<number, CheckableUser>();
+
+      // Find representatives who haven't given ratings
+      missingCustomers.forEach((customer) => {
+        const ratingsForTask = task.ratings
+          .filter((rating) => rating.forCustomer === customer.id)
+          .map((e) => e.createdByUser);
+
+        // Filter out logged in user, so they won't send email to themselves
+        const representatives = customer.representatives?.filter(
+          (user) => user.id !== userInfo?.id,
+        );
+
+        representatives?.forEach((rep) => {
+          if (
+            !ratingsForTask.includes(rep.id) &&
+            !missingRepresentatives.has(rep.id)
+          )
+            missingRepresentatives.set(rep.id, {
+              ...rep,
+              checked: false,
+            });
+        });
+      });
+
+      const missingDevelopers = findMissingDevelopers(
+        task.ratings,
+        allUsers,
+      ).map((dev) => ({ ...dev, checked: false }));
+
+      setMissingUsers([
+        ...Array.from(missingRepresentatives.values()),
+        ...missingDevelopers,
+      ]);
+    }
+  }, [allUsers, customers, task.ratings, task.roadmapId, userInfo]);
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    event.stopPropagation();
+    setIsLoading(true);
+
+    const usersToNotify = missingUsers
+      ?.filter((user) => user.checked)
+      .map((user) => user.id);
+
+    if (usersToNotify) {
+      const res = await dispatch(
+        roadmapsActions.notifyUsers({
+          users: usersToNotify,
+          task,
+          message,
+        }),
+      );
+      setIsLoading(false);
+      if (roadmapsActions.notifyUsers.rejected.match(res)) {
+        if (res.payload) setErrorMessage(res.payload.message);
+        return;
+      }
+      closeModal();
+    }
+  };
+
+  const checkAll = (checked: boolean) => {
+    if (!missingUsers) return;
+    if (!checked) setMissingUsers(previousUsers);
+    if (checked) {
+      const copy = [...missingUsers];
+      setPreviousUsers(copy);
+      setMissingUsers(copy.map((user) => ({ ...user, checked })));
+    }
+    setAllChecked(!allChecked);
+  };
+
+  const checkUser = (checked: boolean, idx: number) => {
+    if (!missingUsers) return;
+    const copy = [...missingUsers];
+    copy[idx].checked = checked;
+    setMissingUsers(copy);
+    setAllChecked(checked && copy.every((user) => user.checked));
+  };
+
+  return (
+    <Form onSubmit={handleSubmit}>
+      <ModalHeader closeModal={closeModal}>
+        <h3>
+          <Trans i18nKey="Send notification by email" />
+        </h3>
+      </ModalHeader>
+      <ModalContent>
+        <p className={classes(css.subtitle)}>
+          <Trans i18nKey="Send the notification to" />
+        </p>
+        <Checkbox
+          label="All"
+          checked={allChecked}
+          onChange={(checked) => checkAll(checked)}
+        />
+        <hr />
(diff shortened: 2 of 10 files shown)


frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,225 @@
...
+        setIsLoading(false);
+        if (roadmapsActions.notifyUsers.rejected.match(res)) {
+          if (res.payload) {
+            setErrorMessage(res.payload.message);
+          }
+        } else {
+          closeModal();
+        }
+      },
+    );
```
IidaKainu, line 146: The readability of this could be improved with async/await
```suggestion
    const res = await dispatch(
      roadmapsActions.notifyUsers({ users: usersToNotify, task }),
    );
    setIsLoading(false);
    if (roadmapsActions.notifyUsers.rejected.match(res)) {
      if (res.payload) setErrorMessage(res.payload.message);
      return;
    }
    closeModal();
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+    const checkedUsers = missingUsers?.filter((user) => user.checked);
+    // Parse RoadmapUsers from list of CheckedUsers
+    const usersToNotify = allUsers?.filter((user) =>
+      checkedUsers?.some((checkedUser) => checkedUser.id === user.id),
+    );
+
+    if (!usersToNotify || usersToNotify?.length === 0) {
+      setIsLoading(false);
+      return setErrorMessage(t('Please select at least one user'));
```
IidaKainu, line 132: Instead of showing an error when no users are selected, the submit button could be disabled.

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,225 @@
...
+          </h3>
+        </ModalHeader>
+        <ModalContent>
+          <p className={classes(css.subtitle)}>
+            <Trans i18nKey="Send the notification to" />
+          </p>
+          <Checkbox
+            label="All"
+            checked={allChecked}
+            onChange={(checked: boolean) => checkAll(checked)}
```
IidaKainu, line 182: The types on the checkbox listeners can be left out due to type inference
```suggestion
            onChange={(checked) => checkAll(checked)}
```

frontend/src/api/api.tsx
```diff
@@ -305,6 +305,11 @@ const patchIntegrationConfiguration = async (
   return response.data as IntegrationConfiguration;
 };
 
+const sendNotification = async (users: RoadmapUser[], task: Task) => {
+  const response = await axios.post(`/notify`, { users, task });
```
IidaKainu, line 309: I'm wondering if the route should follow the pattern of roadmaps/tasks/notify?
```suggestion
  const response = await axios.post(
    `roadmaps/${task.roadmapId}/tasks/${task.id}/notify`,
    { users, task },
  );
```

frontend/src/redux/roadmaps/actions.ts
```diff
@@ -340,3 +340,16 @@ export const patchIntegrationConfiguration = createAsyncThunk<
     }
   },
 );
+
+export const notifyUsers = createAsyncThunk<
+  boolean,
```
IidaKainu, line 345: I think return value should be void. The response status code will tell whether the operation was successful.
```suggestion
  void,
```

IidaKainu (changes requested):
The notified users look correct
The "all" -checkbox behaves really intuitively, great work 👍

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,225 @@
...
+    copy.forEach((_user, idx) => {
+      copy[idx].checked = checked;
+    });
+    setMissingUsers(copy);
+  };
+
+  const checkUser = (checked: boolean, idx: number) => {
+    setAllChecked(false);
+    const copy = [...(missingUsers ?? [])];
+    if (!copy) return;
```
juhoeerola, line 162: `copy` will never be undefined here, so the condition is always false (same above in `checkAll`)

I think I saw the same mistake in some other place as well, outside this pr

```suggestion
    if (!missingUsers) return;
    const copy = [...missingUsers];
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+      // Filter developers, check who haven't given ratings, add checked attribute for all of them
+      const ratingIds = task.ratings.map((rating) => rating.createdByUser);
+      const missingDevelopers = allUsers
+        ?.filter((user) => user.type === RoleType.Developer)
+        .filter((developer) => !ratingIds.includes(developer.id))
+        .map((dev) => ({ ...dev, checked: false }));
+
+      setMissingUsers([
+        ...(missingRepresentatives ?? []),
```
juhoeerola, line 113: won't be undefined
```suggestion
        ...missingRepresentatives,
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,225 @@
...
+          );
+          const ratingsForTask = task.ratings
+            .filter((rating) => rating.forCustomer === customer.id)
+            .map((e) => e.createdByUser);
+
+          if (representatives) {
+            representatives.forEach((rep) => {
+              if (!ratingsForTask.includes(rep.id))
+                if (!missingRepresentatives.some((e) => e.id === rep.id))
+                  missingRepresentatives.push({ ...rep, checked: false });
```
juhoeerola, line 99: `Map<id, user>` could be used, instead of an array, for the deduplication (not that this would be a huge problem any time soon)

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  useEffect(() => {
+    if (getType(userInfo?.roles, task.roadmapId) === RoleType.Admin) {
+      const ratings = task.ratings.map((rating) => ({
+        createdByUser: rating.createdByUser,
+        forCustomer: rating.forCustomer,
+      }));
+      const missingCustomers = customers?.filter((customer) => {
+        const representativeIds = customer?.representatives?.map(
+          (rep) => rep.id,
```
juhoeerola, line 68: Could these use the same functions used for the task listing page? (just filtering out the current user)

juhoeerola (changes requested):

juhoeerola:
> The "all" -checkbox behaves really intuitively, great work 👍

^ this

One possibly nice to have behaviour, would be remembering the individual selection separate to "select all", so that toggling "check all" twice would be idempotent.

Then you would lose the ability to "uncheck all" with that, though... (that would require manual unchecking of every manually checked user)

PatricKangasmaki:
Added all the suggestions!


> Additionally, it might be nice to have ability to include an optional message to the notification, which would be added to the email.

I added the possibility to send messages in theory but didn't add input fields for it yet, I think we should consult our designer on that part.

> One possibly nice to have behaviour, would be remembering the individual selection separate to "all", so that toggling "all" twice would be idempotent.

The checkboxes now save the previous state so toggling back from "all" will have the previously selected ones, big thanks to Iida here!

frontend/src/utils/TaskUtils.ts
```diff
@@ -383,3 +383,33 @@ export const unratedProductOwnerTasks: (
...
+    createdByUser: rating.createdByUser,
+    forCustomer: rating.forCustomer,
+  }));
+  return customers?.filter((customer) => {
+    const representativeIds = customer?.representatives?.map((rep) => rep.id);
+    return !representativeIds?.every((rep) => {
+      return ratings.some((rating) => {
+        if (rating.createdByUser === rep && rating.forCustomer === customer.id)
+          return true;
+        return false;
```
juhoeerola, line 411: Could just return the value of the condition
```suggestion
        return rating.createdByUser === rep && rating.forCustomer === customer.id;
```

frontend/src/utils/TaskUtils.ts
```diff
@@ -383,3 +383,33 @@ export const unratedProductOwnerTasks: (
...
+) => {
+  const ratings = taskRatings.map((rating) => ({
+    createdByUser: rating.createdByUser,
+    forCustomer: rating.forCustomer,
+  }));
+  return customers?.filter((customer) => {
+    const representativeIds = customer?.representatives?.map((rep) => rep.id);
+    return !representativeIds?.every((rep) => {
+      return ratings.some((rating) => {
+        if (rating.createdByUser === rep && rating.forCustomer === customer.id)
```
juhoeerola, line 409: Here also mapping out the `id`s could be skipped
```suggestion
    return !customer.representatives?.every((rep) => {
      return ratings.some((rating) => {
        if (rating.createdByUser === rep.id && rating.forCustomer === customer.id)
```

And `customer` won't be undefined

frontend/src/utils/TaskUtils.ts
```diff
@@ -383,3 +383,33 @@ export const unratedProductOwnerTasks: (
...
 };
+
+export const findMissingDevelopers = (
+  taskRatings: Taskrating[],
+  allUsers: RoadmapUser[],
+) => {
+  const ratingIds = taskRatings.map((rating) => rating.createdByUser);
+  return allUsers
+    ?.filter((user) => user.type === RoleType.Developer)
+    .filter((developer) => !ratingIds.includes(developer.id));
```
juhoeerola, line 394: The filters could be combined, and `allUsers` should not be undefined
```suggestion
    .filter((user) => user.type === RoleType.Developer && !ratingIds.includes(user.id));
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+            </button>
+          </ModalFooterButtonDiv>
+          <ModalFooterButtonDiv>
+            {isLoading ? (
+              <LoadingSpinner />
+            ) : (
+              <button
+                className="button-large"
+                type="submit"
+                disabled={!missingUsers?.filter((user) => user.checked).length}
```
juhoeerola, line 213:
```suggestion
                disabled={!missingUsers?.some((user) => user.checked)}
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+  const checkUser = (checked: boolean, idx: number) => {
+    if (!missingUsers) return;
+    if (allChecked) setAllChecked(false);
+    const copy = [...missingUsers];
+    copy[idx].checked = checked;
+    setMissingUsers(copy);
+  };
+
+  return (
+    <>
```
juhoeerola, line 162: Unnecessary fragment, as the single `Form` element is returned
```suggestion
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+    }
+  };
+
+  const checkAll = (checked: boolean) => {
+    if (!missingUsers) return;
+    if (!checked) setMissingUsers(previousUsers);
+    if (checked) {
+      const copy = [...missingUsers];
+      setPreviousUsers(copy);
+      setMissingUsers(copy.map((_user) => ({ ..._user, checked })));
```
juhoeerola, line 148: The underscore should be dropped, as it usually indicates unused parameter
```suggestion
      setMissingUsers(copy.map((user) => ({ ...user, checked })));
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+      }
+
+      const missingDevelopers = findMissingDevelopers(
+        task.ratings,
+        allUsers,
+      ).map((dev) => ({ ...dev, checked: false }));
+
+      setMissingUsers([
+        ...(Array.from(missingRepresentatives.values()) ?? []),
+        ...(missingDevelopers ?? []),
```
juhoeerola, line 110: Neither will be undefined
```suggestion
        ...Array.from(missingRepresentatives.values()),
        ...missingDevelopers,
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+      allUsers
+    ) {
+      const missingCustomers = findMissingCustomers(task.ratings, customers);
+      const missingRepresentatives = new Map<number, CheckableUser>();
+
+      // Find representatives who haven't given ratings
+      if (missingCustomers) {
+        missingCustomers.forEach((customer) => {
+          // Filter out logged in user, so they won't send email to themselves
+          const representatives = customer?.representatives?.filter(
```
juhoeerola, line 79: also, customer won't be undefined
```suggestion
          const representatives = customer.representatives?.filter(
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+    if (
+      getType(userInfo?.roles, task.roadmapId) === RoleType.Admin &&
+      customers &&
+      allUsers
+    ) {
+      const missingCustomers = findMissingCustomers(task.ratings, customers);
+      const missingRepresentatives = new Map<number, CheckableUser>();
+
+      // Find representatives who haven't given ratings
+      if (missingCustomers) {
```
juhoeerola, line 76: `missingCustomers` won't be false (if the types are correct for `findMissingCustomers`)
```suggestion
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+    }
+    setAllChecked(!allChecked);
+  };
+
+  const checkUser = (checked: boolean, idx: number) => {
+    if (!missingUsers) return;
+    if (allChecked) setAllChecked(false);
+    const copy = [...missingUsers];
+    copy[idx].checked = checked;
+    setMissingUsers(copy);
```
juhoeerola, line 158: A 🍒 on top could be setting `allChecked` if all are checked manually
```suggestion
    setMissingUsers(copy);
    setAllChecked(checked && copy.every((user) => user.checked));
```

frontend/src/components/modals/NotifyUsersModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+            .filter((rating) => rating.forCustomer === customer.id)
+            .map((e) => e.createdByUser);
+
+          if (representatives) {
+            representatives.forEach((rep) => {
+              if (!ratingsForTask.includes(rep.id))
+                if (
+                  !Array.from(missingRepresentatives.values()).some(
+                    (e) => e.id === rep.id,
+                  )
```
juhoeerola, line 92: As far as I could see, this is checking:
```suggestion
                  !missingRepresentatives.has(rep.id)
```

(the nested if conditions could also be combined to one)

juhoeerola (changes requested):
Seems to work great 👍 

Comments mostly on small simplifications
There were also quite a few defensive uses of `?.`, where the type involved didn't include `null` or `undefined`

PatricKangasmaki:
All the suggestions you gave are now fixed and I've added the new input field for the optional message 👍

IidaKainu (approved):
I went over the suggestions and the changes look good, as does the input field. 
Just a small thing I found: the label 'Optional message' is not in translations. Otherwise, I'm giving my approval 👍

juhoeerola (approved):

PR #123, opened 2021-07-29
Use relative value in heatmap

juhoeerola:
Also use hls, as it is nicer to work with.

The specific hsl ranges could be tweaked, if there is some actual design for this at some point.
Now hue range follows relatively closely the previous one.

Couple examples with ~200 tasks, first with random ratings, second with more concentrated (still random) ratings:

200-random

200-concentrated


IidaKainu (approved):
The way you implemented this is neat and the colors fit with figma's color theme. Great work 🔥

PR #125, opened 2021-07-30
Updated & upgraded frontend dependencies

juhoeerola:
- Most notably React 16 -> 17
  - `import React from 'react'` is no longer necessary
- Typefaces were left to current versions.
- Prettier is kept at 2.2.1, as 2.3 introduced formatting difference
- Bootstrap was kept at 4.6, as 5.0 introduced breaking changes


frontend/src/components/TableUnratedTaskRow.tsx
```diff
@@ -153,27 +153,22 @@ export const TableUnratedTaskRow: React.FC<TableTaskRowProps> = ({ task }) => {
...
+                arrow: classes(css.tooltipArrow),
+                tooltip: classes(css.tooltip),
+              }}
+              key={customer.id}
+              title={customer.name}
+              placement="top"
+              arrow
+            >
+              <div className={classes(css.dotContainer)}>
+                <Dot fill={customer.color || 'red'} />
```
IidaKainu, line 168: Looks like this is a leftover from when customer color could be undefined
```suggestion
                <Dot fill={customer.color} />
```

IidaKainu (commented):
I deleted and reinstalled the dependencies. I get the following error on yarn start. 

```
Error: Cannot find module '@babel/code-frame'
Require stack:
- C:\Users\iidak\Documents\VISDOM-Roadmapper\frontend\node_modules\rollup-plugin-terser\index.js
- C:\Users\iidak\Documents\VISDOM-Roadmapper\frontend\node_modules\workbox-build\build\lib\bundle.js
- C:\Users\iidak\Documents\VISDOM-Roadmapper\frontend\node_modules\workbox-webpack-plugin\build\generate-sw.js
- C:\Users\iidak\Documents\VISDOM-Roadmapper\frontend\node_modules\workbox-webpack-plugin\build\index.js
- C:\Users\iidak\Documents\VISDOM-Roadmapper\frontend\node_modules\react-scripts\config\webpack.config.js
- C:\Users\iidak\Documents\VISDOM-Roadmapper\frontend\node_modules\react-app-rewired\scripts\start.js
```

Is this occurring to you?

juhoeerola:
> Is this occurring to you?

No, can't say it is

The `@babel/code-frame` is in `yarn.lock`, so it *should* be installed... did you remove the node_modules directory and then reinstall? I got few other errors without doing that

IidaKainu:
> The @babel/code-frame is in yarn.lock, so it should be installed... did you remove the node_modules directory and then reinstall? I got few other errors without doing that

True: I don't know why this occurs. I removed node_modules and reinstalled with `yarn install` with no luck. 
I also tried deleting the whole project and git cloning it. Installed the dependencies in master and then used `yarn upgrade` on this branch. I get the same error msg still

juhoeerola:
What I did, was:

```sh
cd frontend
git checkout frontend-update-deps
rm -rf node_modules
yarn
```

So, basically I had to remove and reinstall the deps on the upgraded branch, otherwise the old ones would cause problems

PatricKangasmaki (approved):
I managed to get this to build, for me it was just enough to delete node_modules and reinstall them. Since I got in, I tested all the functionalities that I could think of in the app and all things seemed to work fine.
One thing I had to do to get this to compile though: Add rule `'prettier/prettier': ['error', { endOfLine: 'auto' }],` to frontend/.eslintrc.js since it complained about _every single file_ not matching end of line rules (I'm on windows) so I think that would be nice to add for us windows users. Otherwise everything seems good to me 👍

juhoeerola:
>  it complained about _every single file_ not matching end of line rules

The files all have (and should have) LF line endings, so you probably have git changing those locally.

Anyways, I added that

IidaKainu (approved):

PatricKangasmaki (approved):

PR #126, opened 2021-08-02
Type correction

juhoeerola:
Corrected `Roadmap.customers`'s type to include `undefined`

Also removed some unnecessary uses of `?.` and `!.`


PatricKangasmaki (approved):
👍

PR #127, opened 2021-08-06
Redux versions

IidaKainu:
- versions are added under the corresponding roadmap.
- versions can be selected by the id of a roadmap in addition to the selected/current roadmap id

I recall encountering infinite loops on dashboard and planner pages. They do not seem to be a problem anymore, but any reviewers could double-check.
ps, restart frontend with 'yarn start' as the redux modifications will not be used otherwise.


frontend/src/redux/roadmaps/selectors.ts
```diff
@@ -79,3 +79,18 @@ export const chosenIntegrationSelector = (name: string) =>
   createSelector(chosenRoadmapSelector, (roadmap) =>
     roadmap?.integrations.find((it) => it.name === name),
   );
+
+export const roadmapsVersionsSelector = (roadmapId?: number) =>
+  roadmapId
+    ? createSelector(idRoadmapSelector(roadmapId), (roadmap) => {
+        if (!roadmap?.versions) return undefined;
+        return [...roadmap?.versions].sort(
```
juhoeerola, line 87: The condition could be moved to just select the input selector, so the additional processing need not be duplicated in the code.

(Also the `?.` is unnecessary after the check)
```suggestion
    createSelector(roadmapId ? idRoadmapSelector(roadmapId) : chosenRoadmapSelector, (roadmap) => {
        if (!roadmap?.versions) return undefined;
        return [...roadmap.versions].sort(
```

juhoeerola (commented):
Could not reproduce any infinite loops, so those are probably gone. Seems to work fine overall

frontend/src/redux/roadmaps/selectors.ts
IidaKainu: That is really neat - updated to include the changes!

juhoeerola (approved):

PR #128, opened 2021-08-06
Use scaled values in customer stakes

IidaKainu:
- Scaled values are used in customer stakes calculations

Added extra parameter to taskUtils (notWeighted) in case the weighted/scaled values should not be used. (I recall discussion about using scaled values everywhere and adding a convenient on/off switch)


juhoeerola (commented):
👍

IidaKainu:
Fixed, great suggestions 👍

PR #129, opened 2021-08-06
Task rating

IidaKainu:
Task rating is skipped on task creation if there are no customers to rate for

"Fixed" task rating modal styles according to figma
- Well, the styles look correct, but functionality feels messed up. Material UI's rating component allows the user to unselect the rating. Not sure if I would have used the component if I knew the feature could not be disabled...
  - <b>Example</b>: admin represents two customers -> they rate and submit the first customer's rating -> the rating modal can be opened again and the submitted rating can be unselected -> the submission will go through without error msgs, but the rating will not be patched
 
- Any ideas on the user flow? Should all ratings be given in one submission so the modal cannot be revisited? Or perhaps on submission, compare the initial and given ratings for differences instead of having a 'changed' attribute?

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "@material-ui/core": "^4.12.2",
     "@material-ui/icons": "^4.11.2",
+    "@material-ui/lab": "^4.0.0-alpha.60",
     "@reduxjs/toolkit": "^1.6.1",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.0.0",
@@ -41,7 +42,6 @@
     "react-dom": "^17.0.2",
     "react-i18next": "^11.11.3",
     "react-modal": "^3.14.3",
-    "react-rating": "^2.0.5",
     "react-redux": "^7.2.4",
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.3",
--- a/frontend/src/components/RatingBars.module.scss
+++ b/frontend/src/components/RatingBars.module.scss
@@ -1,0 +1,36 @@
+@import '../shared.scss';
+
+.ratingBar {
+  display: flex;
+  align-items: center;
+  gap: 10px;
+
+  .rating {
+    @extend .typography-body-bold;
+  }
+
+  .unrated {
+    @extend .typography-body;
+    color: $COLOR_BLACK40;
+  }
+
+  .icon {
+    padding-right: 17.5px;
+    svg {
+      height: 22.5px;
+      width: 22.5px;
+    }
+  }
+
+  .iconFilled path {
+    fill: $COLOR_AZURE;
+  }
+
+  .iconEmpty path {
+    fill: $COLOR_BLACK10;
+  }
+
+  .iconActive {
+    transform: none;
+  }
+}
--- a/frontend/src/components/RatingBars.tsx
+++ b/frontend/src/components/RatingBars.tsx
@@ -1,12 +1,11 @@
-import { FC } from 'react';
-import Rating from 'react-rating';
+import { FC, useState } from 'react';
+import classNames from 'classnames';
+import { Rating as MaterialRating } from '@material-ui/lab';
 import { TaskRatingDimension } from '../../../shared/types/customTypes';
-import {
-  BusinessValueEmpty,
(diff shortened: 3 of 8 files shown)


PatricKangasmaki:
>Well, the styles look correct, but functionality feels messed up. Material UI's rating component allows the user to unselect the rating. Not sure if I would have used the component if I knew the feature could not be disabled...

Just a thought: maybe it could allow unselecting the rating and actually remove the given rating from data when you unselect and save it (there might be cases where you gave rating but want to remove it for some reason eg. giving rating but changing your mind and needing some time to think or giving rating accidentally before actually knowing the value). 

> Should all ratings be given in one submission so the modal cannot be revisited

In my opinion it's always a good idea to give user the option to change their mind and not require the user to make the ultimate decision instantly.

About styles: Looking good! 👍  One thing I noticed: this is not anything big but to me it looks like the dollar-sign-dots are a bit smaller than in figma. Width seems to be a bit different as well but that's not really noticeable. Left side is current implementation and right side is from figma.

image

IidaKainu:
Corrected the size of the icons. Also now comment box is visible if the modal is reopened and a rating already exists (previously rating needed to change in order for the comment box to appear).

> In my opinion it's always a good idea to give user the option to change their mind and not require the user to make the ultimate decision instantly.

I agree. Then I need to work on the taskrating modal logic a bit

> Just a thought: maybe it could allow unselecting the rating and actually remove the given rating from data when you unselect and save it (there might be cases where you gave rating but want to remove it for some reason eg. giving rating but changing your mind and needing some time to think or giving rating accidentally before actually knowing the value).

What's @juhoeerola 's opinion on this?

juhoeerola:
I'm not sure how the ability to remove ratings would affect the milestone system, as it requires all ratings to be given before task can be added to a milestone. This would at least be difficult from the PO perspective: you can never know how many ratings will be missing from a task tomorrow (e.i. whether or not you can place it into a milestone).

The current system also "prevents" the user from changing any of their ratings after they have given all required ratings for a specific task, as the rated task listing does not provide the rating option.

Then again, the user could just use the backend api directly to edit/delete ratings any time they please...

Overall, the lifecycle of ratings could be defined more clearly (when it is possible to do what with the ratings). And the possible limitations should be enforced in the backend.

IidaKainu:
Finally got the task rating to work as intended thanks to @juhoeerola 
- rating cannot be "unselected" anymore
- submit button is disabled if rating values or comments have not been changed

PatricKangasmaki:
Good job, it seems to work perfectly now 🔥 

I got some thoughts while playing around the task ratings:

- Should you get warning modal when you're about to submit last rating because you can't change your votes after that? I feel like the warning is appropriate when the decision you make will be final.
- Same with clicking the cancel after you have given ratings / wrote comments. Should you get warned that you will lose all the changes? This is not the most important feature since users will most likely know what they're doing when they use cancel but that would be nice quality of life update so you won't lose your changes accidentally.
- Is the `Skip task rating warning modal` still relevant? I found the design in Figma but couldn't find it anywhere in the app

IidaKainu:
> Should you get warning modal when you're about to submit last rating because you can't change your votes after that? I feel like the warning is appropriate when the decision you make will be final.

I remember we discussed about task's lifecycle in a previous weekly - the conclusion was that ratings should be able to be edited. There's just no option for that currently. Thus, I don't think a warning is necessary.

> Same with clicking the cancel after you have given ratings / wrote comments. Should you get warned that you will lose all the changes?

It could be nice to get a warning: I'll ask about it to be sure.

> Is the Skip task rating warning modal still relevant? I found the design in Figma but couldn't find it anywhere in the app

It has not been relevant for a while. However, now that tasks do not need to be fully rated in order to place them into a milestones it could be revisited. We should discuss the topic further in the next weekly.

juhoeerola:
> Should you get warning modal when you're about to submit last rating because you can't change your votes after that? I feel like the warning is appropriate when the decision you make will be final.

I suppose the task detail page would have the rating editing option, so no warnings should be necessary for this.

IidaKainu:
Looks like warnings are unnecessary for task rating. Are there more suggestions on this pr or is it good to go?

PatricKangasmaki:
Seems good to me 👍

PatricKangasmaki (approved):

PR #130, opened 2021-08-09
Move project overview page to higher level router

IidaKainu:
- Changed route from 'roadmap/:selectedroadmapId/overview' to '/overview'.
- The sidebar will no longer show on the overview page.
- chosenRoadmap is removed from Localstorage when moving to the overview page, as no roadmap should be selected.

### DIFF
--- a/frontend/src/components/NavBar.module.scss
+++ b/frontend/src/components/NavBar.module.scss
@@ -16,6 +16,15 @@
   background-color: $COLOR_BLACK5;
 }
 
+.logo {
+  padding-top: 22px;
+  padding-left: 2px;
+  svg {
+    width: 60px;
+    height: 60px;
+  }
+}
+
 .navBarRightSide {
   margin-right: 16px;
   display: flex;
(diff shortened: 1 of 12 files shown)


juhoeerola:
The roadmap in local storage should go away eventually, when the "favourite roadmap" functionality is implemented.
Now the dropdown selector still shows the "selected" roadmap, as it takes it from the state, not from local storage, so the selection should also be cleared from the state.

Also the whole page is scrolling, instead of just the content (like on other pages). It would probably be nicer if the navbar stayed still, both horizontally and vertically. Also the page header could be stationary, so it doesn't disappear when scrolling sideways.

Having the logo in the top left corner, where it would be with the side bar, might also add some nice consistency between the pages.

How it looks  now:
before

How it might look with the changes:
after

IidaKainu:
> the selection should also be cleared from the state

Added! 
Style fixes: 
- the page header  and navbar stays in place
- added visdomLogo 

The vertical scroll looks quite wonky now tough. Also reloading the page shows the incorrect navbar. I'll try to fix these tomorrow.

IidaKainu:
Looks like I forgot requireLogin() from the component. The page reload works as intended now.
I am not sure if I'll be able to do much about the double vertical scrollbar with the current CSS approach.

juhoeerola (changes requested):
The selector widget does not reflect the clearing of the selected roadmap, this should have `else` setting the selection back to the initial placeholder.

https://github.com/Vincit/VISDOM-Roadmapper/blob/93fc9fe4a8ab9c50be5c50a5f68c236572ba2449/frontend/src/components/RoadmapSelectorWidget.tsx#L39-L47

Otherwise looks great 👍 

The double vertical scroll bars are also on the other pages, so that could be a separate issue to clean up. Vertical scroll also shouldn't be as common to have on this page.

juhoeerola (approved):

PR #132, opened 2021-08-10
Replace invalid uses of hooks

juhoeerola:
Noticed that two actions were broken in milestone editor (add / remove task), because invalid hook call errors.

Didn't catch this when reviewing the pr few days back.


PatricKangasmaki (approved):
Good catch 👍

PR #133, opened 2021-08-12
Fix task creation modal

IidaKainu:
- Skip task rating if there are no customers to rate for.

Looks like the task rating draft pr is going to take time to complete. I separated the task creation modal bugfix to this pr.


juhoeerola (approved):
👍

PR #134, opened 2021-08-13
Add tooltip improvements

IidaKainu:
I feel that hovering over the tooltip is more intuitive than clicking on the icon in order for it to stay open. 
- Added interactive prop to tooltip; The tooltip will not close when user hovers over it. The delay to close the tooltip is 500ms.

### DIFF
--- a/frontend/src/components/InfoTooltip.tsx
+++ b/frontend/src/components/InfoTooltip.tsx
@@ -18,6 +18,8 @@ export const InfoTooltip: React.FC<{
       title={title}
       placement="top"
       arrow
+      interactive
+      leaveDelay={500}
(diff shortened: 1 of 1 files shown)


PatricKangasmaki (approved):
This hovering is a smart idea, feels really intuitive! Good job 👍  With the clicking-version, user might have to find the functionality by trial and error since not everyone might get that the icon is clickable right away.
I'd wait until our weekly before merging this though, might be a good idea to demo (or get approval on slack) this to everyone since it's a bit different than the original function.

juhoeerola:
Agreed!

This would also reserve clicking for something else, if needed

PR #135, opened 2021-08-15
Hide task rating buttons if user has no one to represent

PatricKangasmaki:
Fixed bug: `Hide the task rating button(s) if user has no one to represent`

### DIFF
--- a/frontend/src/components/TableUnratedTaskRow.tsx
+++ b/frontend/src/components/TableUnratedTaskRow.tsx
@@ -20,7 +20,7 @@ import {
   allCustomersSelector,
 } from '../redux/roadmaps/selectors';
 import { Dot } from './Dot';
-import { getType } from '../utils/UserUtils';
+import { getType, representsCustomers } from '../utils/UserUtils';
 import css from './TableUnratedTaskRow.module.scss';
 import {
   taskAwaitsRatings,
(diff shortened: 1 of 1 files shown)


frontend/src/components/TableUnratedTaskRow.tsx
```diff
@@ -194,14 +194,16 @@ export const TableUnratedTaskRow: FC<TableTaskRowProps> = ({ task }) => {
...
         )}
-        <button
-          className={classes(css['button-small-filled'])}
-          type="button"
-          disabled={!taskAwaitsRatings(task, userInfo)}
-          onClick={openModal(ModalTypes.RATE_TASK_MODAL)}
-        >
-          <Trans i18nKey="Rate" />
-        </button>
+        {userInfo && representsCustomers(userInfo, task.roadmapId) && (
```
IidaKainu, line 197: This disables rating for developers
```suggestion
        {userInfo &&
          (type === RoleType.Developer ||
            representsCustomers(userInfo, task.roadmapId)) && (
```

IidaKainu (changes requested):
I have to wonder is this bug reproducible? The task should not appear in unrated table if the user cannot rate it 🤔

PatricKangasmaki:
> I have to wonder is this bug reproducible? The task should not appear in unrated table if the user cannot rate it 🤔

I was thinking about admin users in this case, with everyone else that's correct, this won't affect them since the tasks will be in the rated tasks. But with admins (PO): In the current implementation, tasks appear in unrated table until every representative and developer has given ratings even if admin gives their own ratings. The rate button will turn gray after admin gives ratings to all their customers.

The bug I was trying to fix here was instead of displaying the gray button (which implies that admin has given all ratings), the rate button doesn't appear at all if the admin doesn't represent anyone.

Steps to reproduce this:

1. Log in as adminperson1
2. remove adminperson1 from every customers representative-list
3. You can still see the gray rate-buttons implying that you have given ratings (before the fix).

Fix is targeted at the 3. 

In this PR:

3. You can only see notify-button if you don't represent (can't give ratings) any of the customers.

frontend/src/components/TableUnratedTaskRow.tsx
PatricKangasmaki: Good catch 👍  I'll fix that

IidaKainu (approved):
Thanks for the explanation! Works as intended then 👍

PR #136, opened 2021-08-16
Fix developer task rating

IidaKainu:
Developers could not rate unrated tasks (rating modal was empty). Corrected filter predicate.
image


PatricKangasmaki (approved):
This new filter seems to work well! 👍

PR #137, opened 2021-08-16
Add team member

IidaKainu:
Frontend for team member addition

- The modal has two alerts, located below 'create a new link' and 'send link'. I thought it would be make sense to move the alert's next to the 'problems' if there are multiple places where requests can fail. Is this okay?

- Should email field be cleared once the link is successfully sent?


IidaKainu:
(just corrected the typo on admin role warning/caution description)

frontend/src/components/modals/modalparts/TeamMemberModalParts.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { RadioButton } from '../../forms/RadioButton';
+import { RoleType } from '../../../../../shared/types/customTypes';
+import css from './TeamMemberModalParts.module.scss';
+
+const classes = classNames.bind(css);
+
+export const SelectMemberRole: FC<{
+  role: string;
+  onChange: (role: string) => void;
+}> = ({ role, onChange }) => (
```
juhoeerola, line 13: Using the RoleType values could simplify this

```suggestion
export const SelectMemberRole: FC<{
  role: RoleType;
  onChange: (role: RoleType) => void;
}> = ({ role, onChange }) => (
```

frontend/src/components/modals/modalparts/TeamMemberModalParts.tsx
```diff
@@ -0,0 +1,47 @@
...
+        value={RoleType[RoleType.Business]}
+        checked={role === RoleType[RoleType.Business]}
+        onChange={(value: string) => onChange(value)}
+      />
+      <RadioButton
+        label={RoleType[RoleType.Admin]}
+        value={RoleType[RoleType.Admin]}
+        checked={role === RoleType[RoleType.Admin]}
+        onChange={(value: string) => onChange(value)}
+      />
```
juhoeerola, line 36: This could allow taking the options as parameter later, and reduces some of the repetition
```suggestion
      {[RoleType.Developer, RoleType.Business, RoleType.Admin].map((type) => (
        <RadioButton
          key={type}
          label={RoleType[type]}
          value={RoleType[type]}
          checked={role === type}
          onChange={() => onChange(type)}
        />
      ))}
```

frontend/src/utils/string.ts
```diff
@@ -1,5 +1,13 @@
+import { RoleType } from '../../../shared/types/customTypes';
+
 export const titleCase = (s: string): string =>
   s.substr(0, 1).toLocaleUpperCase() + s.substr(1);
 
 export const percent = (maximumFractionDigits?: number) =>
   new Intl.NumberFormat(undefined, { style: 'percent', maximumFractionDigits });
+
+export const getRoleType = (role: string) => {
```
juhoeerola, line 9: This would become unnecessary if the role types aren't handled as strings (other than rendering purposes)

juhoeerola (changes requested):
I was thinking, whether or not the email should stay when toggling the email field. Now it is empty each time the field is opened.

Also, should there be option to unhide the instructions (other than reopen the modal)

The left and right border shadows get clipped, maybe some margin is needed for the div
 
clip

frontend/src/api/api.tsx
```diff
@@ -318,6 +319,22 @@ const sendNotification = async (
...
+const sendInvitation = async (
+  email: string,
+  link: string,
+  role: RoleType,
+  roadmapId: number,
+) => {
+  const response = await axios.post(
+    `roadmaps/${roadmapId}/inviteUser/${link}`,
+    {
+      email,
```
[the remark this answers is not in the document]
IidaKainu: I'm leaning towards the first option, since a link from the server is fetched at this point either way (and the invitation exists in the db).

IidaKainu:
> I was thinking, whether or not the email should stay when toggling the email field

I'd say it's more convenient to have the email stay. (although there might not be many cases where the user starts writing one and closes the field)

> should there be option to unhide the instructions

I am not sure. I added a suggestion on how the option could look, but it is a bit annoying how much space it takes. I'll ask for feedback on slack.


Added all requested changes

IidaKainu:
Modified the add team member modal, as email will be required for the invitation link.

juhoeerola:
If sending by email remains the only option, the modal could have the normal modal footer buttons "Cancel" and "Send link"

frontend/src/components/modals/AddTeamMemberModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+            ) : (
+              <button
+                className={classes(css.linkButton, css.blue)}
+                tabIndex={0}
+                type="button"
+                onClick={() => setOpenInfo(true)}
+              >
+                <Trans i18nKey="Show info" />
+              </button>
+            )}
```
juhoeerola, line 93: This *might* be a bit nicer with conditionally showing the info text, and then the button conditionally selecting its text and toggling the state
```suggestion
            {openInfo && (
              <>
                <b>
                  <Trans i18nKey="Here’s how to add a team member" />
                </b>{' '}
                <Trans i18nKey="Team member addition instructions" />{' '}
              </>
            )}
            <button
              className={classes(css.linkButton, css.blue)}
              tabIndex={0}
              type="button"
              onClick={() => setOpenInfo((prev) => !prev)}
            >
              <Trans i18nKey={openInfo ? 'Hide info' : 'Show info'} />
            </button>
```

juhoeerola (approved):
👍 looks great

PR #138, opened 2021-08-17
Fix CustomerStakesVisualization

IidaKainu:
Add visualization for no ratings (a gray circle with 'No ratings' tooltip)


frontend/src/components/CustomerStakesVisualization.tsx
```diff
@@ -1,18 +1,79 @@
...
+  <Tooltip
+    classes={{
+      arrow: classes(css.tooltipArrow),
+      tooltip: classes(css.tooltip),
+    }}
+    title={title}
+    placement="right"
+    arrow
+  >
+    <div>{children}</div>
```
IidaKainu, line 33: Updated the visualization.

This div is required in order for the tooltip to work (somehow it does not get that StakeVisualizationContent starts with div too). I had to be bit more verbose with CustomerStakesVisualization return jsx due to the div breaking css.

juhoeerola: This seems like a limitation of the tooltip component (https://material-ui.com/components/tooltips/#custom-child-element)

Passing the children directly gives the following warning in the console:
```
index.js:1 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `ForwardRef(Tooltip)`.
```

juhoeerola (approved):

frontend/src/components/CustomerStakesVisualization.tsx
```diff
@@ -1,18 +1,79 @@
...
+const noRatingsStakes = [
+  { id: 1, value: 0.5 },
+  { id: 2, value: 0.3 },
+  { id: 3, value: 0.1 },
+  { id: 4, value: 0.1 },
+];
+
+const StakeVisualizationTooltip: FC<{
+  title: string;
+  children: any;
```
juhoeerola, line 22: Almost forgot to mention: the `children` is typed by `FC`, so overriding that with less strict type could be dropped
```suggestion
```

PR #140, opened 2021-08-20
Fix modals

IidaKainu:
Since roadmap overview was moved higher level router, roadmap deletion modal did not open. Moved modal opening to MainRouter.

Fixed EditVersionModal's styles.


juhoeerola (approved):
👍 good catch

PR #141, opened 2021-08-20
Fix

juhoeerola:
Fixed issues:
- the new task table cells could extend outside the given space
- misplaced `key` property in list of elements


IidaKainu (approved):
👍

PR #142, opened 2021-08-20
Task util refactor

juhoeerola:
- started to rewrite / reorganise task utils into (hopefully) reusable predicates
- rewrote sorting to
  - use decorate-sort-undecorate automatically for calculated keys
  - allow object property key as sorting key
  - use `undefined` as `NO_SORT`
- added `useSorting()` hook & SortingArrow component

### DIFF
--- a/frontend/src/utils/SortUtils.ts
+++ b/frontend/src/utils/SortUtils.ts
@@ -21,7 +21,7 @@ type Key<T, K> = PropertyKey<T, K> | ((value: T) => K);
 
 type Comparison<T> = (a: T, b: T) => number;
 
-export type Sort<T, K = any> =
+export type SortBy<T, K = any> =
   | undefined
   | {
       key: Key<T, K>;
@@ -32,7 +32,7 @@ export type Sort<T, K = any> =
 const sortKeyCompare = <T, K>(
   key: Key<T, K>,
   compare: Comparison<K>,
-): Sort<T, K> => ({ key, compare });
+): SortBy<T, K> => ({ key, compare });
 
 export const sortKeyNumeric = <T>(key: Key<T, number>) =>
   sortKeyCompare(key, (a, b) => a - b);
@@ -41,7 +41,7 @@ export const sortKeyLocale = <T>(key: Key<T, string>) =>
   sortKeyCompare(key, (a, b) => a.localeCompare(b));
 
 export const sort = <T, K>(
-  by: Sort<T, K>,
+  by: SortBy<T, K>,
   order: SortingOrders = SortingOrders.ASCENDING,
 ) => {
   // no sorting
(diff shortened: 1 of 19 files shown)


frontend/src/utils/SortUtils.ts
```diff
@@ -1,27 +1,89 @@
...
+// If optional property is needed for sorting, the key must be a function
+type PropertyKey<T, K> = {
+  [P in keyof T]: T[P] extends K ? P : never;
+}[RequiredKeys<T>];
+
+type Key<T, K> = PropertyKey<T, K> | ((value: T) => K);
+
+type Comparison<T> = (a: T, b: T) => number;
+
+export type Sort<T, K = any> =
```
IidaKainu, line 24: Would the name 'SortBy' suit this type better? Currently there's a sort function and Sort type, which may be a bit confusing

```suggestion
export type SortBy<T, K = any> =
```

frontend/src/utils/TaskUtils.ts
```diff
@@ -288,15 +236,29 @@ export const calcWeightedTaskPriority = (task: Task, roadmap: Roadmap) => {
...
-  rating: Taskrating,
-) => rating.createdByUser === user.id;
+export const awaitsUserRatings = (
+  user: UserInfo,
+  roadmap: number | Roadmap,
+) => {
+  const roadmapId = typeof roadmap === 'number' ? roadmap : roadmap.id;
+  const type = getType(user.roles, roadmapId);
+  if (type === RoleType.Admin || type === RoleType.Business) {
+    const reps = user.representativeFor?.filter(
```
IidaKainu, line 246: Clients or customers would describe the value better
```suggestion
    const clients = user.representativeFor?.filter(
```

frontend/src/utils/TaskUtils.ts
```diff
@@ -305,134 +267,40 @@ export const isRatedByCustomer = (
...
+      : awaitsUserRatings(user, roadmap);
   }
+
   if (user.type === RoleType.Admin || user.type === RoleType.Business) {
-    const represented = customers?.filter((customer) =>
-      customer.representatives?.find((rep) => rep.id === user.id),
-    );
-    return !!represented?.find(
-      (customer) => !task.ratings.some(isRatedByCustomer(customer, user)),
+    const represented = roadmap.customers?.filter((customer) =>
```
IidaKainu, line 283: Maybe clients or customers would suit better here too (at least I thought 'wait who are represented' for a second)
```suggestion
    const clients = roadmap.customers?.filter((customer) =>
```

IidaKainu (changes requested):
Really neat refactor! Although generics complicate reading of the implementations (took a while to understand😅), the solutions are 100% improvements.

I have only a few naming suggestions, otherwise this is good to go 🔥

juhoeerola:
Thanks, those are arguably better names

Added the suggestions

IidaKainu (approved):

PR #143, opened 2021-08-23
Default selected roadmap

PatricKangasmaki:
Selecting default roadmap can be used in frontend and backend now. 
I'm keeping this as a draft for now since I've got some open questions about a few functionalities:

-Is it okay that users have: `defaultRoadmapId: null` if they haven't selected any roadmap to be their default. Updating the default roadmap then happens with patch  /users/:id.

-What would be a good way to handle situations where you delete the roadmap that has been selected by some users as their default? For example, Markku has selected roadmap 1 as his default roadmap. Roadmap 1 gets deleted -> what should happen to the field `defaultRoadmapId` on Markku? Currently there's a check after logging in where defaultRoadmapId is compared to all the roadmaps in store. If there's a match -> redirect to the matched roadmap. If there's no match -> patch the user to have `defaultRoadmapId: null` and redirect to the overview page. Currently this approach has a disadvantage where users will still have the missing roadmap as their default one until they log in and the state of defaultRoadmapId will not update until page is refreshed.  Would it be better if the deletion functionality was in the backend or maybe roadmap deletion modal (or should the current implementation just be polished a bit to remove the disadvantages)?


juhoeerola:
Having the default roadmap nullable would probably be the best option, as it may not be defined.

Handling the deletion would then be simplest using the db constraints: default roadmap is foreign key referencing roadmaps table, and set null on delete.

PatricKangasmaki:
> Handling the deletion would then be simplest using the db constraints: default roadmap is foreign key referencing roadmaps table, and set null on delete.

Added this functionality, it works now exactly as needed. Thank you!

frontend/src/components/ProjectSummary.tsx
```diff
@@ -14,52 +14,70 @@ import { isCustomer } from '../utils/CustomerUtils';
...
-      </Link>
-    </div>
-  </Popover>
-);
+}> = ({ roadmapId, userId, anchorEl, open, onClose }) => {
+  const dispatch = useDispatch<StoreDispatchType>();
+
+  const handleButton = async () => {
+    dispatch(
+      userActions.patchDefaultRoadmap({
```
juhoeerola, line 35: Could maybe use `await`, or drop the `async`

frontend/src/redux/user/types.ts
```diff
@@ -7,6 +7,7 @@ export interface UserInfo {
   id: number;
   roles: RoadmapRole[];
   representativeFor?: Customer[];
+  defaultRoadmapId: number;
```
juhoeerola, line 10: This should include null
```suggestion
  defaultRoadmapId: number | null;
```

server/src/migrations/20210816092729_addDefaultRoadmapColumnToUser.ts
```diff
@@ -0,0 +1,19 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  return knex.schema.table('users', (table) => {
+    table
+      .integer('defaultRoadmapId')
+      .unsigned()
```
juhoeerola, line 7:
```suggestion
    table
      .integer('defaultRoadmapId')
      .unsigned()
      .nullable()
```

server/src/migrations/20210816092729_addDefaultRoadmapColumnToUser.ts
```diff
@@ -0,0 +1,19 @@
...
+      .references('id')
+      .inTable('roadmaps')
+      .onDelete('SET NULL')
+      .index();
+  });
+}
+
+export async function down(knex: Knex): Promise<void> {
+  return knex.schema.table('users', (table) => {
+    table.dropForeign(['defaultRoadmapId']);
```
juhoeerola, line 17: The actual column needs also to be dropped
```suggestion
    table.dropForeign(['defaultRoadmapId']);
    table.dropColumn('defaultRoadmapId');
```

juhoeerola (commented):
Have I understood correctly, the default roadmap can't be unset from the ui? (other than: create new -> select as default -> delete)

The popover menu could at least not give "set as default" option for the current default. Or maybe it could have option to remove the default selection.

PatricKangasmaki:
Added all your suggestions! I refactored the popover menu a bit: now the old "set as a default" button is rendered conditionally. If you open menu of a selected project, it will display "deselect default project" and set defaultRoadmapId to null. And if the project is not selected, it will display the original "select as a default"

frontend/src/pages/HomePage.tsx
```diff
@@ -5,30 +5,34 @@ import { roadmapsSelector } from '../redux/roadmaps/selectors';
...
 
-    const match = roadmaps.find((map) => map.id === previousRoadmap.id);
+  if (userInfo.defaultRoadmapId) {
+    const match = roadmaps.some((map) => map.id === userInfo.defaultRoadmapId);
     if (match) {
       return (
-        <Redirect to={`${paths.roadmapHome}/${previousRoadmap.id}/dashboard`} />
+        <Redirect
+          to={`${paths.roadmapHome}/${userInfo.defaultRoadmapId}/dashboard`}
+        />
```
IidaKainu, line 29:
```suggestion
        <Redirect
          to={`${paths.roadmapHome}/${userInfo.defaultRoadmapId}${paths.roadmapRelative.dashboard}`}
        />
```

frontend/src/pages/HomePage.tsx
```diff
@@ -5,30 +5,34 @@ import { roadmapsSelector } from '../redux/roadmaps/selectors';
...
       );
     }
   }
 
-  if (!roadmaps) return <LoadingSpinner />;
-
   return roadmaps.length > 0 ? (
-    <Redirect to={`${paths.roadmapHome}/${roadmaps[0].id}/dashboard`} />
+    <Redirect to="/overview" />
```
IidaKainu, line 35:
```suggestion
    <Redirect to={paths.overview} />
```

frontend/src/pages/HomePage.tsx
```diff
@@ -5,30 +5,34 @@ import { roadmapsSelector } from '../redux/roadmaps/selectors';
...
   ) : (
     <Redirect to="/getstarted" />
```
IidaKainu, line 37:
```suggestion
    <Redirect to={paths.getStarted} />
```

frontend/src/redux/user/index.ts
```diff
@@ -1,5 +1,12 @@
 import { createSlice } from '@reduxjs/toolkit';
-import { getUserInfo, login, logout, register } from './actions';
+// To do: Import default roadmap action
```
IidaKainu, line 2: Is this comment a leftover? 🤔

IidaKainu (changes requested):
Looks good 👍 
I noticed a few places where paths could be used in place of hard coded values

PatricKangasmaki:
Added the suggested paths! And yeah, that comment was some old leftover from when I was planning how to proceed, removed it now 😅

IidaKainu (approved):

PR #145, opened 2021-08-24
Adjust frontend to permission model

IidaKainu:
Many api routes would give 403 forbidden errors, when user is not admin.
Now api calls are made only if user has the required permission.


juhoeerola (approved):
👍

PR #147, opened 2021-08-26
Send invitation link by email

juhoeerola:
todo:
- set up correct env. variables for aws
- some design for the emails


PatricKangasmaki (approved):
Looking good 🔥 I think this current email is just fine as a placeholder until we get the new design for it 👍

juhoeerola:
Moved aws access keys out from the required env, as those are provided from other source in staging

PR #151, opened 2021-08-27
Remove local planner weights

IidaKainu:
Since only admins see the planner page and only few might be present in a roadmap, the edition of user weights is saved instantaneously.

### DIFF
--- a/frontend/src/pages/PlannerWeightsPage.tsx
+++ b/frontend/src/pages/PlannerWeightsPage.tsx
@@ -1,38 +1,25 @@
 import { useEffect, useState } from 'react';
 import { Alert } from 'react-bootstrap';
 import { shallowEqual, useDispatch, useSelector } from 'react-redux';
-import { Trans, useTranslation } from 'react-i18next';
+import { Trans } from 'react-i18next';
 import classNames from 'classnames';
-import InfoIcon from '@material-ui/icons/InfoOutlined';
 import { Slider } from '../components/forms/Slider';
 import {
-  plannerCustomerWeightsSelector,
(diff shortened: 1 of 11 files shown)


IidaKainu:
- rebased the branch on top of master
- removed the tooltip from the planner weights page as the description is already written on the page (I can also move the info into the tooltip if it's preferable)

frontend/src/pages/PlannerWeightsPage.tsx
```diff
@@ -56,79 +48,33 @@ export const PlannerWeightsPage = () => {
...
-
-  useEffect(() => {
-    setNoChanges(
-      !customerWeights.some(({ customerId, weight }) => {
-        const saved = customers?.find(({ id }) => id === customerId)?.weight;
-        return weight !== saved;
-      }),
+  const handleSliderChange = async (customerId: number, weight: number) => {
+    setLocalCustomers(
+      localCustomers!.map((customer) =>
```
juhoeerola, line 53: It's better to provide update function to set state, when the previous state needs to be referenced,
this way e.g. batched updates work reliably
```suggestion
    setLocalCustomers((previous) =>
      previous!.map((customer) =>
```

juhoeerola (changes requested):
Looks great 👍 

just a couple suggestions

IidaKainu:
Added the suggestions 👍 

> provide update function to set state, when the previous state needs to be referenced

Feels a bit surprising I have missed such a fundamental part about the useState hook. I need to do some documentation reading 😄

juhoeerola (approved):

PR #152, opened 2021-08-27
Add task relations model

juhoeerola:
Endpoints: get, post, delete.
Patch doesn't really make so much sense for this.

The relations are returned as a list of directed edges, with the nodes `from` and `to`, and the `type` of the relation.

### DIFF
--- a/server/src/api/taskrelation/taskrelation.controller.ts
+++ b/server/src/api/taskrelation/taskrelation.controller.ts
@@ -1,0 +1,40 @@
+import { RouteHandlerFnc } from '../../types/customTypes';
+import Task from '../tasks/tasks.model';
+import { TaskRelation } from './taskrelation.model';
+
+export const getRelations: RouteHandlerFnc = async (ctx) => {
+  const tasks = (
+    await Task.query().where({
+      roadmapId: Number(ctx.params.roadmapId),
+    })
+  ).map((t) => t.id);
+
+  ctx.body = await TaskRelation.query()
+    .whereIn('from', tasks)
+    .orWhereIn('to', tasks);
+};
+
+const tasksInRoadmap = async (roadmapId: number, taskIds: number[]) => {
+  const tasks = await Task.query().where({ roadmapId }).whereIn('id', taskIds);
+  return tasks.length === taskIds.length;
+};
+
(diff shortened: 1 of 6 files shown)


server/src/api/taskrelation/taskrelation.controller.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const tasksInRoadmap = async (roadmapId: number, taskIds: number[]) => {
+  const tasks = await Task.query().where({ roadmapId }).whereIn('id', taskIds);
+  return tasks.length === taskIds.length;
+};
+
+export const addRelation: RouteHandlerFnc = async (ctx) => {
+  const { from, to, type } = ctx.request.body;
+  // assuming that from and to are not the same
+  if (await tasksInRoadmap(Number(ctx.params.roadmapId), [from, to])) {
```
PatricKangasmaki, line 25: Should this have error check for the case where **from** and **to** have the same value? That way it could give an informative error (currently it says: 400 - bad request). Might save from some confusion in the future

PatricKangasmaki (commented):
Looking good and seemed to work correctly 👍 One part caught my attention, nothing big though

server/src/api/taskrelation/taskrelation.controller.ts
juhoeerola: 400 is definitely the correct response code
Adding specific error message for that might be nice, although the frontend should probably not allow relating task to itself in the first place

PatricKangasmaki: Yeah I think the error code is good 👍 
> although the frontend should probably not allow relating task to itself in the first place

That's also true. ~~I think that having a message in the error like 'Task cannot be related to itself' could help in a case where a developer who's working on frontend haven't dived into the backend yet. Seeing just "bad request" might cause initial confusion without knowing what's actually wrong with the request.~~

But now that I'm writing this and thinking of these scenarios, I think that in relating tasks it is kinda obvious that you relate them to other tasks and not to theirselves. So leaving it like this is probably more than fine

juhoeerola: Added an explicit check for `from !== to`, so it is clearer when reading the code

PR #153, opened 2021-08-31
Retain search context between login/register pages

IidaKainu:
- redirectTo context used to get lost when moving to register page
- now the joining link can be used with registering

### DIFF
--- a/frontend/src/pages/RegisterPage.tsx
+++ b/frontend/src/pages/RegisterPage.tsx
@@ -4,9 +4,9 @@ import { withStyles } from '@material-ui/core/styles';
 import { Alert } from 'react-bootstrap';
 import classNames from 'classnames';
 import { Trans, useTranslation } from 'react-i18next';
-import { useDispatch } from 'react-redux';
+import { shallowEqual, useDispatch, useSelector } from 'react-redux';
 import Checkbox from '@material-ui/core/Checkbox';
-import { Link, useHistory, useLocation } from 'react-router-dom';
+import { Link, useLocation, Redirect, matchPath } from 'react-router-dom';
 import { userActions } from '../redux/user';
 import { ModalContent } from '../components/modals/modalparts/ModalContent';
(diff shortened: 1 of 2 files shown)


PatricKangasmaki:
Works well! 👍 One side-effect I found: If you have some existing page in your url when registering (/overview or /roadmaps/  etc..), you will skip the `It seems you don’t have any projects yet...` -page. Might not be a problem because there's not really many reasons for a new user to have something extra in their url except maybe when using shared computer. I just wanted to point this out now so it will not come as a surprise later on.

Getting redirected to this page as a first-time user might be really confusing though if it were to happen
image

Would it mess things up or be really burdensome to implement if anything else but _join project_ -links would redirect to the `It seems you don’t have any projects yet...` -page? If it's quick and won't cause more side-effects, it might be worth implementing.

frontend/src/pages/RegisterPage.tsx
```diff
@@ -126,7 +127,7 @@ export const RegisterPage = () => {
     if (userActions.register.rejected.match(res)) {
       errorHandler(res.payload?.response?.data);
     } else if (userActions.register.fulfilled.match(res)) {
-      history.push('/login');
+      history.push(paths.loginPage + search);
```
juhoeerola, line 130: I think this would be the place to check the redirect, if that would be done.

Also, the path for history, or the `to` property for links, can be provided as an object:
```suggestion
      history.push({ pathname: paths.loginPage, search });
```

IidaKainu:
> One side-effect I found: If you have some existing page in your url when registering (/overview or /roadmaps/ etc..), you will skip the It seems you don’t have any projects yet... -page

Great feedback! Added the following:

> anything else but join project -links would redirect to the It seems you don’t have any projects yet... -page

juhoeerola (approved):
👍 seems to work well

PatricKangasmaki (approved):

PR #154, opened 2021-08-31
Fix no ratings CustomerStakesVisualization

IidaKainu:
- Small visual fix: removed the first grey dot

### DIFF
--- a/frontend/src/components/CustomerStakesVisualization.tsx
+++ b/frontend/src/components/CustomerStakesVisualization.tsx
@@ -11,10 +11,9 @@ import colors from '../colors.module.scss';
 const classes = classNames.bind(css);
 
 const noRatingsStakes = [
-  { id: 1, value: 0.5 },
-  { id: 2, value: 0.3 },
-  { id: 3, value: 0.1 },
-  { id: 4, value: 0.1 },
+  { id: 1, value: 0.3 },
+  { id: 2, value: 0.23 },
+  { id: 3, value: 0.2 },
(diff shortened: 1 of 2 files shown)


frontend/src/components/CustomerStakesVisualization.tsx
```diff
@@ -11,7 +11,6 @@ import colors from '../colors.module.scss';
 const classes = classNames.bind(css);
 
 const noRatingsStakes = [
-  { id: 1, value: 0.5 },
   { id: 2, value: 0.3 },
   { id: 3, value: 0.1 },
   { id: 4, value: 0.1 },
```
PatricKangasmaki, line 16: Would it be better if the values would get gradually lower? Currently the id: 2 is the biggest dot, and 3 & 4 are the same size.

In my opinion something like:

```
const noRatingsStakes = [
  { id: 2, value: 0.3 },
  { id: 3, value: 0.25 },
  { id: 4, value: 0.1 },
];
```

looked a bit better. What do you think?

IidaKainu: I submitted one version with the second value of 0.23 (I felt 0.25 was too large).
Overall I thought both look good, so I sent a question to Minna and your suggestion is now also design-approved.

PatricKangasmaki: Looking good 👍

PatricKangasmaki (approved):

PR #156, opened 2021-09-08
Skip server tests on frontend-only changes

juhoeerola:
The current tests only test the server, so it is unnecessary (and also misleading) to run those on frontend changes.

The only benefit from running tests each time is that brittle tests are more likely noticed.


PatricKangasmaki (approved):
That's a smart idea 👍

PR #157, opened 2021-09-08
Use real ses client only on production environment

juhoeerola:
Otherwise the sent email is only logged to the console.

This should allow easier experimenting on the development environment.

Set `NODE_ENV=production` to use the real service.


PatricKangasmaki (approved):
👍

PR #160, opened 2021-09-20
Email verification

juhoeerola:
Verification link is automatically sent when:
- new user is registered
- user's email is updated

Verified email is now required for all pages requiring login, except:
- user info page, this page is very WIP, but it now shows the
  verification status, it has a button to send a new verification link,
  and changing the user's email would also be done here
- verify email page, obviously

Verified email is now also required on the server side.

### DIFF
--- a/frontend/src/pages/VerifyEmailPage.tsx
+++ b/frontend/src/pages/VerifyEmailPage.tsx
@@ -17,15 +17,17 @@ export const VerifyEmailPage = requireLogin(({ userInfo }) => {
   }>();
 
   useEffect(() => {
-    if (!verificationId || userInfo.emailVerified) return;
+    if (userInfo.emailVerified) {
+      history.push(paths.getStarted);
+      return;
+    }
+    if (!verificationId) return;
 
     setIsLoading(true);
     dispatch(userActions.verifyEmail({ user: userInfo, verificationId }))
       .then((res) => {
         if (userActions.verifyEmail.rejected.match(res)) {
           setErrorMessage(res.payload?.message ?? 'Something went wrong');
-        } else {
-          history.push(paths.getStarted);
         }
       })
       .finally(() => setIsLoading(false));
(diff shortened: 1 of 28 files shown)


frontend/src/pages/VerifyEmailPage.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (userInfo.emailVerified) {
+      history.push(paths.getStarted);
+      return;
+    }
+    if (!verificationId) return;
+
+    setIsLoading(true);
+    dispatch(userActions.verifyEmail({ user: userInfo, verificationId }))
+      .then((res) => {
+        if (userActions.joinRoadmap.rejected.match(res)) {
```
IidaKainu, line 29: No error messages currently show up
```suggestion
        if (userActions.verifyEmail.rejected.match(res)) {
```

server/src/api/roadmaps/roadmaps.routes.ts
```diff
@@ -18,46 +18,51 @@ import customerRouter from '../customer/customer.routes';
...
+  prefix: '/roadmaps',
+});
 
 roadmapRouter.use(requireAuth);
-roadmapRouter.use('/roadmaps/:roadmapId', requireRole);
+roadmapRouter.use('/:roadmapId', requireRole);
 
-roadmapRouter.get('/roadmaps', getRoadmaps);
-roadmapRouter.post('/roadmaps', postRoadmaps);
+roadmapRouter.get('/', getRoadmaps);
```
IidaKainu, line 28: Why get roadmaps -route does not require verified email? I do not see why an unverified account would need access to this route, since they are not in any roadmap

IidaKainu (changes requested):
I finally got around fully reviewing this pr. Overall, this is solid work 🔥 
I noticed only one line, that needs changes.

Ps, should we create test case(s) for the requireVerifiedEmail -middleware? This could be done as a separate pr

server/src/api/roadmaps/roadmaps.routes.ts
juhoeerola: Existing account's email may also become unverified by changing it, or possibly it could be set as unverified if it starts to generate bounces.

Also it otherwise gets stuck on a loading spinner on home page, before getting to the user info page.

IidaKainu: Thanks for explaining, looks like this is a well thought decision

IidaKainu (approved):

PR #165, opened 2021-10-05
Fix unmounted state update error

AleksiKytola:
Dropdown.tsx was giving console warnings about unmounted state updates due to it containing links that navigate to other pages, and not being rendered anymore and its onClick handler tried to call setOpen(false) on it.
Now the document onclick listener is removed in useEffect cleanup handler if component is unmounted


juhoeerola (approved):
Seems fine 👍

PR #170, opened 2021-10-07
Make waiting tasks clickable so they can be opened

JimiRantanen:


frontend/src/components/TaskTableUnrated.tsx
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable jsx-a11y/click-events-have-key-events */
 /* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
juhoeerola, line 2: Unrelated to the changes, these seem to be left over from earlier and could be removed before merging.

PR #172, opened 2021-10-11
Move Clients list to its own page

IidaKainu:
Admins and business-people can access the page.
List all clients in case of admin, otherwise list represented clients.
Only admins can add clients.


AleksiKytola (changes requested):
Looking good apart from a few little things:

- The clients page does not have a header/title at thetop, like task and people pages do.
- As a side note to the above, I noticed that the people pages' title is not positioned/styled in the same position as the tasks pages title. Noticeable when swapping between these pages. Perhaps one unified css class should be used for these page titles.

AleksiKytola (changes requested):
A good base for the TopBar. Easy to review well divided small commits :)
Now that I'm looking at it I'm thinking we should make it just take a callback to run when the "add button" is clicked instead of just making it open a modal.
Then the interface is encapsulated a bit better I think and the parent component has full control of what to do when the button is pressed.
Exactly like your onSearchChange prop works currently.

PR #174, opened 2021-10-11
Fix bug: "100vw" widths to "100%" to avoid horizontal scrollbar

ghost:
Previously a horizontal scrollbar appeared because "100vw" width does not account for when a vertical scrollbar appears, constraining the width and causing a horizontal one to appear as well


IidaKainu (approved):
👍

PR #176, opened 2021-10-14
Remove privatekey from integration response

juhoeerola:
The private key is only needed to be sent from the frontend for adding
or updating. There is no reason to send it back.

Otherwise the private key would be visible to all the admins in the roadmap.


AleksiKytola:
Good job spotting this one

PR #177, opened 2021-10-14
Client overview

IidaKainu:
Added an Overview-component, which can be used in task, client etc. overviews. 
The component will require some modification once @JimiRantanen has merged task overview edit buttons, but it should be doable.

- Client table rows are not fully clickable like task rows. I'd say this should be done in a different pr as adding a link messes up the row styles.


AleksiKytola (commented):
Overview component seems like a nice thing to have. The only thing I have to add is that its api is relatively complex that its not necessarily self-documenting enough at this point, I guess dropping a quick 
```
// ** 
 * JsDoc
 * https://jsdoc.app/about-getting-started.html
 * /
```
inline documentation could help future us remember how to use it

IidaKainu:
Added documentation. Hope it suffices, there seems to be multiple ways of writing JSDoc for React components.

Ps, got the client table rows fully clickable - I'll submit it in a different pr

PR #187, opened 2021-10-29
Improve react-flow styling

PatricKangasmaki:
- Added new files for edges and connection lines to components folder
- Fixed bug where edges would be behind task-components
- Added custom styling for edges and connectionlines (gradient works now)
- Fixed bug where edges would be invisible if they were straight lines (sourceY and targetY same value)
- Edges have hover effect


IidaKainu (approved):
Looking good 👍

PR #188, opened 2021-11-01
Improve task-map styles

IidaKainu:
Small style changes to task-map

### DIFF
--- a/frontend/src/components/TaskRatingsText.module.scss
+++ b/frontend/src/components/TaskRatingsText.module.scss
@@ -15,11 +15,3 @@
     margin-left: 11.75px;
   }
 }
-
-.doneIcon {
-  color: $COLOR_EMERALD;
-  margin-right: 5px;
-  &.white {
-    color: $COLOR_WHITE;
-  }
-}
--- a/frontend/src/components/TaskRatingsText.module.scss
+++ b/frontend/src/components/TaskRatingsText.module.scss
@@ -15,3 +15,11 @@
     margin-left: 11.75px;
   }
 }
+
+.doneIcon {
+  color: $COLOR_EMERALD;
+  margin-right: 5px;
+  &.white {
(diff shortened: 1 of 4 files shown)


PatricKangasmaki:
Looking great! 👍  One thing I'd probably change is the location of the checkmark. I would prefer it to be there right away even before the text for it to be consistent with the style of task-tables. Also, user will most likely always read the task name first so they would instantly be able to see that "oh this task is completed". Maybe there could also be the lighter text-color on completed tasks like in task maps. One negative I could see is that the task names wouldn't be lined up as nicely though.
This is just my personal preference, I'm open for discussion and any ideas!

image

PatricKangasmaki:
> Looking great! 👍 One thing I'd probably change is the location of the checkmark. I would prefer it to be there right away even before the text for it to be consistent with the style of task-tables. Also, user will most likely always read the task name first so they would instantly be able to see that "oh this task is completed". Maybe there could also be the lighter text-color on completed tasks like in task maps. One negative I could see is that the task names wouldn't be lined up as nicely though. This is just my personal preference, I'm open for discussion and any ideas!
> 
> image

Made a quick sketch (in paint, my favorite design tool available) about that, don't know anymore. Does it look a bit off or is it just me? Maybe it's the orange and green being next to each other

image

IidaKainu:
> Does it look a bit off or is it just me?

Maybe it's the way the icon misaligns the text. In either option, there's a lot of info and icons in such a small area, so it might be a factor. I'll test a few versions and post on slack.

IidaKainu:
I tested Patric's suggestion and completed tasks are definitely faster to notice (also Minna prefers this one). I left out the grey text, I think the tasks looks a bit too complicated with it.

PatricKangasmaki (approved):
Looks good! 🔥

PR #189, opened 2021-11-03
Add info icon to taskmap

PatricKangasmaki:
Added info-icon next to the center-map icon in Task map. Info text currently gives tips about deleting relations.


IidaKainu (changes requested):
The position for the info icon looks good

Found a few things

PatricKangasmaki:
Added all your suggestions, you had good ideas! 👍 And now the info-icon is using relative position 😄

IidaKainu (approved):

PR #191, opened 2021-11-03
Block timeline

juhoeerola:
At the moment this scrolls the selected element into view, but that does not work when synchronising the two views on /planner/graph page. Well, it would without 'smooth' scroll, but then it doesn't feel nice.

### DIFF
--- a/frontend/src/components/BlockGraph.tsx
+++ b/frontend/src/components/BlockGraph.tsx
@@ -1,4 +1,4 @@
-import { ReactNode } from 'react';
+import { RefObject, ReactNode } from 'react';
 import classNames from 'classnames';
 import css from './BlockGraph.module.scss';
 
@@ -56,6 +56,7 @@ interface BlockViewProps<T> {
     height: number;
   }) => ReactNode;
   className?: string;
+  innerRef?: RefObject<HTMLDivElement>;
 }
 
 type BlockGraphProps<T> = {
@@ -75,9 +76,10 @@ export function BlockView<T>({
   origin = 0,
   children,
   className,
+  innerRef,
 }: BlockViewProps<T>) {
   return (
-    <div className={classes(css.viewContainer, className)}>
+    <div ref={innerRef} className={classes(css.viewContainer, className)}>
       <div className={classes(css.viewItems)}>
         {sizing(items.slice(origin), dimensions, limits).map((props, idx) =>
           children({ ...props, index: origin + idx }),
--- a/frontend/src/components/BlockGraph.tsx
+++ b/frontend/src/components/BlockGraph.tsx
@@ -1,0 +1,138 @@
+import { ReactNode } from 'react';
+import classNames from 'classnames';
+import css from './BlockGraph.module.scss';
+
+const classes = classNames.bind(css);
+
+const lerp = (values: number[], lo: number, hi: number) => {
+  const min = Math.min(...values);
+  const max = Math.max(...values);
+  return values.map(
+    (x) => lo + Math.floor(((x - min) / (max - min)) * (hi - lo)),
+  );
+};
+
+interface DimensionLimits {
+  minWidth: number;
+  maxWidth: number;
+  minHeight: number;
+  maxHeight: number;
+}
+
+type DimensionFn<T> = (
+  item: T,
+  idx: number,
+) => { width: number; height: number };
+
+function sizing<T>(
+  items: T[],
+  dimensions: DimensionFn<T>,
+  limits: DimensionLimits,
+) {
+  const sizes = items.map(dimensions);
+  const w = lerp(
+    sizes.map((x) => x.width),
+    limits.minWidth,
+    limits.maxWidth,
+  );
+  const h = lerp(
+    sizes.map((x) => x.height),
+    limits.minHeight,
+    limits.maxHeight,
+  );
+  return items.map((item, i) => ({ item, width: w[i], height: h[i] }));
+}
+
+/* eslint-disable react/require-default-props */
+interface BlockViewProps<T> {
+  items: T[];
+  origin?: number;
+  dimensions: DimensionFn<T>;
+  limits: DimensionLimits;
+  children: (props: {
+    item: T;
+    index: number;
+    width: number;
+    height: number;
+  }) => ReactNode;
+  className?: string;
+}
+
+type BlockGraphProps<T> = {
+  title: ReactNode;
+  xLabel: ReactNode;
+  yLabel: ReactNode;
+  selected: number;
+  setSelected: (_: number) => void;
+  id: (item: T) => string | number;
+  children: (props: { item: T; selected: boolean }) => ReactNode;
+} & Omit<BlockViewProps<T>, 'children'>;
+
+export function BlockView<T>({
+  items,
+  dimensions,
+  limits,
+  origin = 0,
+  children,
+  className,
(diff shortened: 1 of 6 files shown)


frontend/src/components/BlockGraph.tsx
```diff
@@ -0,0 +1,156 @@
+import { RefObject, ReactNode } from 'react';
+import classNames from 'classnames';
+import css from './BlockGraph.module.scss';
+
+const classes = classNames.bind(css);
+
+const lerp = (values: number[], lo: number, hi: number) => {
```
PatricKangasmaki, line 7: I'd add a bit more descriptive names for the `hi` and `lo` to improve readability

frontend/src/pages/RoadmapGraphPage.tsx
```diff
@@ -46,6 +45,33 @@ export const RoadmapGraphPage = () => {
...
   );
 
+  // TODO: scrolling selected into view from BlockGraph doesn't work with this
+  const a = useRef<HTMLDivElement>(null);
+  const b = useRef<HTMLDivElement>(null);
+  const refs = useRef([a, b]);
+  const scrolling = useRef(0);
+
+  useEffect(() => {
+    const handleScroll = (e: any) => {
```
PatricKangasmaki, line 55: `(e: any)`

Is there any good ways to give this one a type instead of any? I was struggling with the same in my code, couldn't figure out any ways to fix it.

frontend/src/pages/RoadmapGraphPage.tsx
```diff
@@ -61,79 +87,74 @@ export const RoadmapGraphPage = () => {
...
-                    {ver.name}
-                  </p>
-                </div>
-              );
-            })}
+      <BlockGraph
+        title={
+          <div className={classes(css.titleContainer)}>
+            <h2 className={classes(css.graphTitle)}>{t('workValueTitle')}</h2>
+            <InfoTooltip title={t('tooltipMessage')}>
```
PatricKangasmaki, line 116: I think we should adding these tooltips messages soon, it could be a good to have them before the demo gets out. Our designer might have good ideas for the contents of these.

PatricKangasmaki (commented):
Changes are looking good! 👍

One thing caught my attention:
image

When there's a lot of milestones, the page will get cluttered by scrollbars. Is there any way to make this a bit cleaner (while still maintaining the visual cues that they are all scrollable?

AleksiKytola:
> Changes are looking good! +1
> 
> One thing caught my attention: image
> 
> When there's a lot of milestones, the page will get cluttered by scrollbars. Is there any way to make this a bit cleaner (while still maintaining the visual cues that they are all scrollable?

I was thinking that its enough that the top view is scrollable, and the bottom view has the scrollbar hidden and just keeps in sync with the top view

frontend/src/pages/RoadmapGraphPage.tsx
juhoeerola: I don't know any good way either. The type for `scroll` event handler on div seems to be `(e: Event) => any` or `(this: HTMLDivElement, e: Event) => any`

And `Event` isn't generic over the target element, also `React.SyntheticEvent<T>` doesn't compile.

I think the cleanest way here would be type assertion on the event target ( and I should be using the `currentTarget` )
```ts
    const handleScroll = (e: Event) => {
      // ...
      const target = e.currentTarget as HTMLDivElement | null;
      if (target) {
        const other = target === a.current ? b : a;
        window.requestAnimationFrame(() =>
          other.current?.scroll({ left: target.scrollLeft }),
        );
      }
    };
```
or picking the correctly typed object based on the event target
```ts
      const target = e.currentTarget;
      const self = (target === a.current ? a : b).current;
      const other = (target === a.current ? b : a).current;
      if (self && other) {
        window.requestAnimationFrame(() =>
          other.scroll({ left: self.scrollLeft }),
        );
      }
```


One way the *types would match* would be...
```ts
    function handleScroll(this: HTMLDivElement, e: Event) {
      // ...
      const other = this === a.current ? b : a;
      window.requestAnimationFrame(() =>
        other.current?.scroll({ left: this.scrollLeft }),
      );
    }
    const elements = [a.current, b.current];
    elements.forEach((el) => el?.addEventListener('scroll', handleScroll));
```

frontend/src/components/BlockGraph.tsx
juhoeerola: naming is hard, spelt out the names now, but I'm open for suggestions if there are some clearly better ones

AleksiKytola:
For me, the view doesn't scroll to the selected block horizontally. The page only scrolls vertically to the block view component itself, but the block views "internal" horizontal scroll doesnt adjust so that the block is visible.

juhoeerola:
> For me, the view doesn't scroll to the selected block horizontally.

This is due to the synchronisation with the other, and because the timeline has `behaviour: smooth` on the scroll.
It seem like triggering horizontal scroll on the other prevents the original `scrollIntoView` from finishing.
Instantaneous scroll worked fine though.

https://github.com/Vincit/VISDOM-Roadmapper/pull/191/files#diff-4a512f36c597da27e59164fb3a7018c40bb875a5723494ce7114b4abb17d0e71R48

AleksiKytola:
```
              ref={
                index === selected
                  ? (e) => {
                      if (e) {
                        e.parentElement?.parentElement?.scrollTo({
                          top: e.offsetTop,
                          left:
                            e.offsetLeft -
                            e.parentElement.parentElement.offsetLeft,
                          behavior: 'smooth',
                        });
                        console.log(`Scrolling to elem ${e.offsetLeft}`);
                      }
                    }
                  : undefined
              }
```

This one works for horizontal scrolling, even when synced. I dont know why the original instantenous scrolling worked for you, when I did a bit of googling it seemed that the element.scrollIntoView doesnt even handle horizontal scrolling in the first place

AleksiKytola:
visdom-scroll

juhoeerola:
> This one works for horizontal scrolling, even when synced. I dont know why the original instantenous scrolling worked, when I did a bit of googling it seemed that the element.scrollIntoView doesnt even handle horizontal scrolling in the first place

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameters the `inline` parameter suggests that it would also handle horizontal positioning...

Finding the correct parent element is simple though, as the `BlockView` takes inner ref that will be set to the containing div.
Not sure if chaining the refs like that is good though (and not sure how to handle ref object)
```ts
        <BlockView
          innerRef={(e) => {
            if (e) ref.current = e;
            if (typeof innerRef === 'function') innerRef(e);
          }}
          {...viewProps}
          className={classes(css.graphInner)}
        >
          {({ item, index, width, height }) => (
            <div
              ref={
                index === selected
                  ? (e) => {
                      if (e) {
                        ref.current?.scrollTo({
                          top: e.offsetTop,
                          left: e.offsetLeft - ref.current.offsetLeft,
                          behavior: 'smooth',
                        });
                      }
                    }
                  : undefined
              }
```

frontend/src/components/BlockGraph.tsx
juhoeerola: Yeah, `mapToRange` might be a better name

as a side note: we "could" use `map`, it's just a method name, but that would be confusing

AleksiKytola (approved):

PatricKangasmaki (approved):

PR #192, opened 2021-11-03
Add functionality for removing task relations through task-map

PatricKangasmaki:
First version of relation deleting. Currently there's two ways to delete a relation: right clicking an edge or hovering over it and pressing backspace or delete.

There will be additional features added to deleting relations but this is enough currently. The rest of the features will be added after automatic layout generator has been implemented, that will give us more info how things should work.

### DIFF
--- a/frontend/src/api/api.tsx
+++ b/frontend/src/api/api.tsx
@@ -23,6 +23,7 @@ import {
   GetRoadmapBoardLabelsRequest,
   Version,
   VersionRequest,
+  TaskRelation,
 } from '../redux/roadmaps/types';
 import { IntegrationBoard, Integrations } from '../redux/types';
 import {
@@ -366,6 +367,17 @@ const addTaskRelation = async (relation: TaskRelationRequest) => {
   return response.status === 200;
 };
 
+const removeTaskRelation = async (
+  roadmapId: number,
+  relation: TaskRelation,
+) => {
+  const response = await axios.delete(
+    `/roadmaps/${roadmapId}/tasks/relations`,
+    { data: relation },
+  );
+  return response.status === 200;
+};
+
 export const api = {
   getRoadmaps,
   addRoadmap,
@@ -409,4 +421,5 @@ export const api = {
   verifyEmail,
   sendEmailVerificationLink,
   addTaskRelation,
+  removeTaskRelation,
 };
--- a/frontend/src/components/TaskMapEdge.tsx
+++ b/frontend/src/components/TaskMapEdge.tsx
@@ -1,11 +1,51 @@
-import { FC } from 'react';
+import { FC, MouseEvent, useState, useEffect, useCallback } from 'react';
+import { useDispatch } from 'react-redux';
 import classNames from 'classnames';
 import { getMarkerEnd, getBezierPath } from 'react-flow-renderer';
+import { StoreDispatchType } from '../redux';
+import { roadmapsActions } from '../redux/roadmaps';
 import css from './TaskMapEdge.module.scss';
 
 const classes = classNames.bind(css);
 
 const DrawPath: FC<any> = ({ id, d, markerEnd }) => {
+  const [selected, setSelected] = useState<string | undefined>(undefined);
+  const dispatch = useDispatch<StoreDispatchType>();
+
+  // Should be given 'id' as param in the form: 'from-taskId-to-taskId'
+  const deleteRelation = useCallback(
+    async (tbdeleted: string) => {
+      const data = tbdeleted.split('-');
+      await dispatch(
+        roadmapsActions.removeTaskRelation({
+          from: Number(data[1]),
+          to: Number(data[3]),
+          type: 0,
+        }),
+      );
+      dispatch(roadmapsActions.getRoadmaps());
+    },
+    [dispatch],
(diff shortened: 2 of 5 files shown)


frontend/src/components/TaskMapEdge.tsx
```diff
@@ -1,11 +1,52 @@
...
+        deleteRelation(selected);
+        setSelected(undefined);
+      }
+    };
+    document.addEventListener('keydown', handleBackspace);
+    return () => {
+      document.removeEventListener('keydown', handleBackspace);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [deleteRelation, dispatch]);
```
juhoeerola, line 48: Should this include the `deleteRelation` as dependency? It could be wrapped in a `useCallback`

frontend/src/components/TaskMapEdge.tsx
```diff
@@ -25,8 +66,9 @@ const DrawPath: FC<any> = ({ id, d, markerEnd }) => {
         className={`react-flow__edge-path ${classes(css.invisiblePath)}`}
         d={d}
         markerEnd={markerEnd}
-        // eslint-disable-next-line no-alert
-        onClick={() => alert(`To do: remove connections = ${id}`)}
+        onContextMenu={handleDelete}
```
juhoeerola, line 69:
```suggestion
        onAuxClick={handleDelete}
        onContextMenu={(e) => { e.preventDefault(); }}
```

The context menu seems to trigger after mouse down, not after a click (down + up), using the `onAuxClick` event might be nicer. Then just check if the mouse button is the right one (pun intended)

juhoeerola (changes requested):

PatricKangasmaki:
Added all your suggestions!

- Wrapped deleteRelation in useCallback()
- Changed context menu to aux-click, this was a really good change since it now leaves room for extra features later (left click etc.) 👍
- Blocked context menu from opening on the task-map canvas. This could be changed to be blocked only on the edges, but it felt annoying when the menu kept opening if you missclicked and didn't hit the egde.
- keyCode -> key

juhoeerola (approved):
Looking good 👍

frontend/src/redux/roadmaps/actions.ts
```diff
@@ -532,3 +532,15 @@ export const addTaskRelation = createAsyncThunk<
...
   }
 });
+
+export const removeTaskRelation = createAsyncThunk<
+  boolean,
+  TaskRelationRequest,
+  { rejectValue: AxiosError }
+>('removeTaskRelation', async (relationRequest, thunkAPI) => {
+  try {
+    return await api.removeTaskRelation(relationRequest);
```
IidaKainu, line 542: Changing the TaskRelationRequest to TaskRelation would simplify the task-map logic a bit
I have removed the type from addTaskRelation in another branch, my bad for implementing it this way in the first place
```suggestion
export const removeTaskRelation = createAsyncThunk<
  boolean,
  TaskRelation,
  { rejectValue: AxiosError }
>('removeTaskRelation', async (relationRequest, thunkAPI) => {
  try {
    const roadmapId = chosenRoadmapIdSelector(
      thunkAPI.getState() as RootState,
    )!;
    return await api.removeTaskRelation(roadmapId, relationRequest);
```

frontend/src/api/api.tsx
```diff
@@ -366,6 +366,14 @@ const addTaskRelation = async (relation: TaskRelationRequest) => {
   return response.status === 200;
 };
 
+const removeTaskRelation = async (relation: TaskRelationRequest) => {
+  const response = await axios.delete(
+    `/roadmaps/${relation.roadmapId}/tasks/relations`,
+    { data: { from: relation.from, to: relation.to, type: relation.type } },
```
IidaKainu, line 372: Is there a reason the relation is nested like this?
```suggestion
    { from: relation.from, to: relation.to, type: relation.type },
```
Or if the TaskRelationRequest type is changed to TaskRelation
```suggestion
    relation,
```

IidaKainu (changes requested):
A few thoughts, looks good otherwise 👍

frontend/src/api/api.tsx
PatricKangasmaki: Axios.delete requires you to nest data like that if you want to give it body params. I cleaned it up a bit though, now it has ` { data: relation },` instead of `{ data: { from: relation.from, to: relation.to, type: relation.type } },`

_axios.delete does support a request body. It accepts two parameters: url and optional config. You can use config.data to set the request body and headers as follows: ..._
https://stackoverflow.com/questions/51069552/axios-delete-request-with-body-and-headers

frontend/src/redux/roadmaps/actions.ts
PatricKangasmaki: Changed it to be like this 👍 Cleaned up the id's of the edges as well from `from-:taskId-to-:taskId-in-:roadmapId` to `from-:taskId-to-:taskId` since the roadmap gets selected in the actions

IidaKainu (approved):
> Axios.delete requires you to nest data like that if you want to give it body params.

Ah, good to know 👍 Looking good

PR #193, opened 2021-11-04
Visual fixes to Overview pages and EditableText

AleksiKytola:


frontend/src/components/EditableText.module.scss
```diff
@@ -15,9 +15,12 @@
   @extend .layout-row;
   @extend .flexGrow;
   align-items: center;
+  text-align: start;
 
   padding-left: 21px;
-  padding-bottom: 2px;
+  margin-top: 14px;
+  margin-bottom: 14px;
+  padding-right: 21px;
```
PatricKangasmaki, line 23: You could shorten these to:

`margin: 14px 0px;`
`padding: 0px 21px`

PatricKangasmaki (changes requested):
This is looking great! 👍 There were a few things that I found:

If I recall correctly, there was some talk in weekly some time ago about not displaying the title twice (example in pic). Should it be changed or is this okay?

image

Some text overflow checks would be nice add as well, they're currently breaking badly (Editing works well, the textfield expands and you can write on multiple lines):

image

**EDIT:**

Just noticed that the overflow is only a problem when the words are too long. Otherwise it seems to work nicely 👍 It would be good to address this corner case anyways. I think it would be better if the absurdly long words would be under the other elements / have invisible overflow / be on multiple lines like other texts if this were to happen instead of going over them and making it look messy

juhoeerola:
Some combination of
```css
  overflow: hidden;
  white-space: nowrap;
  text-overflow: ellipsis;
  max-width: 40ch;
```
could work nicely for the text.

The ellipsis doesn't work with `display: flex;` though, so if that would be used it would need inner container for the text.

AleksiKytola:
image
Added overflow: hidden to avoid text overflowing. Do note that the text only overflows if its a long text with no spaces interrupting since the word-wrap property of the element only allows the text to wrap between words. So in a typical use case it works as expected

AleksiKytola:
I didn't remove the duplicate title yet, as I think it looks fine this way and the only memory I have about removing duplicates was about removing the color dot from one of the fields in the customer overview

frontend/src/pages/ClientOverviewPage.tsx
```diff
@@ -77,7 +77,7 @@ const ClientOverview: FC<{
       {
         label: t('Client'),
         keyName: 'clientDotName',
-        value: clientDotName,
+        value: <>{client.name}</>,
```
IidaKainu, line 80: The option for jsx is mostly for styling purposes, so in this case the fragment can be left out
```suggestion
        value: client.name,
```

IidaKainu (commented):
Sorry for not noticing the review request earlier. Found a minor thing to change, otherwise looks good 👍

PR #195, opened 2021-11-10
Retain edited text on error

juhoeerola:
- When there is an error the `cancel` button only clears the error
  message and does not close the edit field
- The text is only cleared after successful submit

Alternative for #196


juhoeerola:
The other alternative was chosen

PR #196, opened 2021-11-10
Retain edited text on error

juhoeerola:
- Used the common form field component for displaying the error.
- Added a little gap between the columns to keep those from overlapping
  with the edit field border.
- The text field is focused when moving to edit mode. (The focus leaves
  the cursor at the start.)

Alternative for #195


IidaKainu (approved):
Looks good to me 👍

PR #197, opened 2021-11-11
Fix rated tasks

juhoeerola:
This should fix the "rated" logic for admins.
For admins I also show the tasks that are missing rating from others, otherwise the "notify" option would be unavailable.


IidaKainu (approved):
Works as intended and the implementation is neat 👍

PR #203, opened 2021-11-15
Add frontend build step to CI

AleksiKytola:


juhoeerola (approved):
Seems to work fine, we can separate the server/frontend ci jobs later if need arises

PR #204, opened 2021-11-18
Fix import modal dropdown

juhoeerola:
Didn't find a way to grow the modal conveniently to fit the opened list. However, attaching the list to the document body instead of the modal allows it to go over the boundaries of the modal window such that it is visible and usable.

Also small visual fix on the task page top bar buttons.


IidaKainu (approved):
The dropdown list looks good like this to me 🔥

PatricKangasmaki (approved):
👍

PR #205, opened 2021-11-19
Taskmap Drag n Drop

IidaKainu:
- Disabled task group dragging for now as it conflicts with task drag-n-drop (the whole group gets dragged on task drag). A fix can be developed as a seperate pr


PatricKangasmaki (changes requested):
This looks and feels really good 🔥 A few things caught my eye:

- If you drag a task and go over relation-edges, the cursor will turn into pointer (to display that you can delete the relation) and you can even delete the relations while have a task in your hand. I think it should be disabled during dragging if that is not too big thing to implement
- While you are dragging a task outside of any current synergy-groups, there's nothing to give user a hint that you can drop the task anywhere to create a new synergy (where there's just that one task alone). Maybe some visual cue or adding some information to the info-button would be nice.

IidaKainu:
Really valuable suggestions: added fixes for all of them 👍 

Visual cue for task drag on outside of any groups could be nice. I added a box-shadow for an example, but it is not very visible with selected tasks (blue task background). Any suggestions on better styling?

frontend/src/components/TaskMapTask.tsx
```diff
@@ -0,0 +1,160 @@
...
+          taskId={taskId}
+          selected={selected}
+          setSelectedTask={setSelectedTask}
+          snapshot={snapshot}
+          provided={provided}
+          toChecked={toChecked}
+          fromChecked={fromChecked}
+          disableDragging={disableDragging}
+        />
+      );
```
juhoeerola, line 157:
```suggestion
      const element = (
        <SingleTask
          taskId={taskId}
          snapshot={snapshot}
          provided={provided}
          disableDragging={disableDragging}
          {...rest}
        />
      );
      return snapshot.isDragging
        ? ReactDOM.createPortal(element, document.getElementById('taskmap')!)
        : element;
```

Might be worth deduplicating the element construction, as it has quite a few props

The props that are only forwarded could also be passed without explicitly naming each

frontend/src/pages/TaskMapPage.tsx
```diff
@@ -188,63 +105,141 @@ export const TaskMapPage = () => {
...
+  };
+
+  const onDragEnd = async (result: DropResult) => {
+    const { source, destination, draggableId } = result;
+    const draggedTaskId = Number(draggableId);
+    const backupList = copyRelationList(taskRelations);
+    const copyList = copyRelationList(taskRelations);
+
+    try {
+      if (!destination) await onDragMoveOutside(draggedTaskId, copyList);
```
juhoeerola, line 191: There are a few unnecessary copies, e.g. `onDragMoveOutside()` effectively does it's own copy with `map` & `filter`s,
and the `backupList` could be avoided if the possibly failing operations are done before setting the new state.

This was just an observation and no need to change this now

frontend/src/pages/TaskMapPage.tsx
```diff
@@ -188,63 +105,141 @@ export const TaskMapPage = () => {
...
     dispatch(roadmapsActions.getRoadmaps());
   };
 
+  const addSynergyRelations = async (from: number, to: number[]) => {
+    const res = await dispatch(
+      roadmapsActions.addSynergyRelations({ from, to }),
+    );
+    if (roadmapsActions.addSynergyRelations.rejected.match(res))
+      throw new Error();
+  };
```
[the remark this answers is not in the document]
IidaKainu: thanks, nice to learn about unwrap 👍

frontend/src/pages/TaskMapPage.tsx
IidaKainu: Thanks for notifying about the multiple copies, I did not notice it
I wanted to temporarily set the new lists to the state so the animations look smooth (otherwise the task jumps around). Not sure how it's possible without an backup

IidaKainu:
Thanks for the through review: fixes added

juhoeerola (approved):
Looking good 👍

PatricKangasmaki (approved):
> I added a box-shadow for an example, but it is not very visible with selected tasks (blue task background). Any suggestions on better styling?

I think the box shadow looks good and is enough for now 👍 If it seems to be confusing / not seen later, it's easy to fix

PR #211, opened 2021-11-29
Add redux state sync

IidaKainu:
https://trello.com/c/zdkcdyKu/385-redux-staten-synkronointi-v%C3%A4lilehtien-v%C3%A4lill%C3%A4-redux-persist
- Initialize new tabs with existing redux state from other tabs
- Sync redux state across tabs in realtime

I also fiddled with redux persist in redux-sync-persist -branch. It allows saving the state to localstorage. However, persist does not fit in with the current implementation since UserState defines if the user is logged in.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -49,6 +49,7 @@
     "react-window": "^1.8.6",
     "recharts": "^2.0.10",
     "redux": "^4.1.0",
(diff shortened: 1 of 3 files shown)


PatricKangasmaki (approved):
nice 👍

PR #214, opened 2021-12-07
Infrastructure-as-code

AleksiKytola:
Adds AWS CDK deployment scripts and necessary launch scripts for server, continuous delivery action for github.


juhoeerola (commented):
Looking good overall

PR #215, opened 2021-12-08
New user page

JimiRantanen:
A new user  page, where editing any info or trying to delete the account prompts for the user's password.

Mostly ready, with some known issues:

- [x] - Tests fail, because users PATCH and DELETE routes now require the user's current password - they need to be rewritten
- [ ] - EditableText needs a property to clear text when opening edit
- [ ] - EditableText needs a property to control input type (password in this case)
- [x] - No redirect to login / front page when deleting user, state is left intact

Tests fixed, maybe split the rest to separate tickets?


AleksiKytola:
Im fine with the editabletext issues being fixed in another ticket:
 - EditableText needs a property to clear text when opening edit
 - EditableText needs a property to control input type (password in this case)

However the last one seems like a part of this one and should be included
 - No redirect to login / front page when deleting user, state is left intact

JimiRantanen:
All right, I will get to it next week

frontend/src/redux/user/actions.ts
```diff
@@ -58,6 +63,26 @@ export const register = createAsyncThunk<
...
+  boolean,
+  UserModifyRequest,
+  { rejectValue: AxiosError }
+>('/user/patchUser', async (userPatch: UserModifyRequest, thunkAPI) => {
+  try {
+    if (userPatch.deleteUser) {
+      if (await api.deleteUser(userPatch)) {
+        return true;
+      }
+    } else if (await api.patchUser(userPatch)) {
```
juhoeerola, line 76:
```suggestion
      return await api.deleteUser(userPatch));
    }
    if (await api.patchUser(userPatch)) {
```

Would be clearer to just return the result directly, since the `false` path doesn't do anything more

juhoeerola (commented):
Few suggestions

juhoeerola:
> EditableText needs a property to ...

The editable text could maybe pass through props to the underlying input element more transparently, so user could just provide any that the input can use.

frontend/src/components/UserInfoCard.tsx
```diff
@@ -1,16 +1,162 @@
...
+  const dispatch = useDispatch<StoreDispatchType>();
+  const [sent, setSent] = useState(false);
+  const [sending, setSending] = useState(false);
+
+  const ProjectRow: TableRow<RoadmapRole> = ({ item: roadmapRole, style }) => {
+    const { roadmapId, type } = roadmapRole;
+    const roadmap = useSelector(idRoadmapSelector(roadmapId));
+    return (
+      <div className={classes(css.virtualizedTableRow)} style={style}>
+        <td className="styledTd roleIcon">
```
juhoeerola, line 32:
```suggestion
        <div className="styledTd roleIcon">
```

Noticed that this was complaining about incorrect nesting of elements, div should do fine here.

juhoeerola (changes requested):

frontend/src/components/UserInfoCard.tsx
```diff
@@ -1,16 +1,157 @@
...
+        <div className={classes(css.textContainer)}>
+          <EditableTextWithButtons
+            onOk={emailOnOk}
+            value={userInfo.email}
+            fieldId="email"
+            format=""
+          />
+          {!userInfo.emailVerified ? (
+            <p className={classes(css.unVerifiedText)}>(unverified)</p>
+          ) : null}
```
IidaKainu, line 107:
```suggestion
          {!userInfo.emailVerified && (
            <p className={classes(css.unVerifiedText)}>(unverified)</p>
          )}
```

frontend/src/components/UserInfoCard.tsx
```diff
@@ -1,16 +1,157 @@
...
+        </div>
+        {!userInfo.emailVerified && (
+          <button
+            className="button-small-outlined"
+            type="button"
+            onClick={resendOnClick}
+            disabled={sent || sending}
+          >
+            Resend verification email
```
IidaKainu, line 116: There's a few translations missing in this component

IidaKainu (changes requested):
Found a few minor things, otherwise functions and looks good to me 👍

juhoeerola:
Added fixes and resolved conflicts with master

IidaKainu (approved):

PR #217, opened 2021-12-09
Make sidebar buttons scrollable

juhoeerola:
Previously the buttons would stretch the page vertically beyond the
viewport height, which would add extra scrollbars to the page and make
the scrolling overall a bad experience.


PatricKangasmaki (approved):
Looking good 👍 Was there any updates on the small indicator that the sidebar is scrollable (like the small arrow) or did the idea get dropped?

IidaKainu:
> Was there any updates on the small indicator that the sidebar is scrollable (like the small arrow) or did the idea get dropped?

A separate design card was added for the indicator 👍

PR #220, opened 2021-12-16
Add stretching for long names in taskmap

IidaKainu:
https://trello.com/c/z9xskpGu/389-fiksaa-tyylit-mukautumaan-pitkiin-nimiin

### DIFF
--- a/frontend/src/components/TaskMapTask.module.scss
+++ b/frontend/src/components/TaskMapTask.module.scss
@@ -2,8 +2,9 @@
 
 .singleTask {
   @extend .layout-row;
-  width: 309px;
-  height: 41px;
+  width: 360px;
+  min-height: 40px;
+  max-height: 90px;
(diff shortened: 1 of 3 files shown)


PatricKangasmaki (approved):
Looking a lot better and more clear now! 👍 One thing I noticed that if there are really long words, they will overflow and be hidden and the next word will go on the second row. This makes it a bit confusing (on the other hand, I don't think there are actually any words that would be that long so this isn't a realistic situation).

image


Anyways, it could be fixed with `word-wrap: break-word` in the taskname styling but I'm not sure if it may cause cutting off words in a situations where it shouldn't 🤔 At least in basic situations it worked just as wanted. You can test if it is actually needed and decide whether to add it or not!

image

IidaKainu:
Looks like `word-wrap: break-word` did not affect much else except for this corner case you found, so I added it to the implementation 🙏

PR #221, opened 2021-12-16
Fix overview browse buttons

IidaKainu:
https://trello.com/c/97muxs4h/410-eteenp%C3%A4in-taaksep%C3%A4in-napit-el%C3%A4%C3%A4-kun-kelaa-taskista-toiseen-riippuen-taskin-nimen-pituudesta

### DIFF
--- a/frontend/src/components/Overview.module.scss
+++ b/frontend/src/components/Overview.module.scss
@@ -13,6 +13,10 @@
   width: 100%;
   gap: 24px;
 
+  & > *:not(.buttons) {
+    margin-top: 12px; // the default padding of the buttons
+  }
+
   .name {
     @extend .layout-row;
     align-items: center;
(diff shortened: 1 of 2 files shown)


PatricKangasmaki (changes requested):
Currently if the title is a one long word, it will push the buttons out of screen.

image

For example, adding these lines of code to the `.name` class in `Overview.module.scss` would cause the overflow to break on multiple lines.

```
max-width: 70%; // Can be adjusted to some other value, this just looked good to me
display: inline-block;
overflow-wrap: break-word;
```

Result:

image

IidaKainu:
> Currently if the title is a one long word, it will push the buttons out of screen.

Great catch, added

PatricKangasmaki (approved):

PR #222, opened 2021-12-16
Show selected colour in customer edit modal

juhoeerola:
The colour is shown as a dot next to the hex input field.

dot

### DIFF
--- a/frontend/src/components/Dot.tsx
+++ b/frontend/src/components/Dot.tsx
@@ -1,9 +1,12 @@
 import { FC } from 'react';
 
-export const Dot: FC<{ fill: string }> = ({ fill }) => (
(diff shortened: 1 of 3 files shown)


IidaKainu (approved):
Looks good to me 👍 
I was discussing this card with Patric and he interpreted it a bit differently so let's request his opinion too

PatricKangasmaki (approved):
>I was discussing this card with Patric and he interpreted it a bit differently so let's request his opinion too

Originally I thought that the testers meant that they would like to see the color _in text format_ as well, but it wouldn't really make any sense since you can choose literally any color you want, not just the basic predefined ones (like red green blue purple etc.). It would be impossible to define things like dark purple, darker purple, purple with even more darkness, and things between them etc. as text so your implementation is great imo.

PR #224, opened 2021-12-17
Use role translations

IidaKainu:
https://trello.com/c/xbtRuXDg/418-muuta-teksteihin-business-rooli-sales-representative

### DIFF
--- a/frontend/src/components/modals/JoinProjectModal.tsx
+++ b/frontend/src/components/modals/JoinProjectModal.tsx
@@ -53,7 +53,7 @@ export const JoinProjectModal: Modal<ModalTypes.JOIN_PROJECT_MODAL> = ({
               i18nKey="Join project info 1/2"
               values={{
                 name: invitation.roadmap!.name,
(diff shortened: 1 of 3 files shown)


PatricKangasmaki (approved):
👍
One thing to note: there's now a lot of things called `business` in the code and comments while business person got changed to sales representative.  Should everything be called for example `sales` in the code for consistency or is it better to use that time for more important work since all the developers in the project will probably be fine with it being called business. Maybe it could be brought up in next meeting?

IidaKainu:
That's an understandable concern. This is likely not the best solution, but at least 'sales representative' is isolated to the translations 🤔

PatricKangasmaki:
>at least 'sales representative' is isolated to the translations

I think this solution is good for now. It's better to change everything everywhere after there's a clear guideline how it should be done. Imo this is ready to merge, just remember to create a new ticket about the business-type in the code and bring it up in next meeting after the holidays 👍

PR #225, opened 2021-12-21
Add auto-layout to taskMap

IidaKainu:


frontend/src/pages/TaskMapPage.tsx
```diff
@@ -68,47 +69,77 @@ export const TaskMapPage = () => {
...
+        type: 'special',
+        sourcePosition: Position.Right,
+        targetPosition: Position.Left,
+        draggable: false,
+        // dagre coordinates are in the center, calculate top left corner
+        position: { x: node.x - node.width / 2, y: node.y - node.height / 2 },
+        data: {
+          label: (
+            <TaskGroup
+              listId={Number(id)}
```
juhoeerola, line 109:
```suggestion
              listId={id}
```

The `id` could be kept as a string, as it is now just converted back and forth.

frontend/src/pages/TaskMapPage.tsx
```diff
@@ -68,47 +69,77 @@ export const TaskMapPage = () => {
...
+              taskIds={synergies.sort((a, b) => a - b)} // FIXME: ordering prevents render bugs
```
juhoeerola, line 110:
```suggestion
              taskIds={synergies}
```
I couldn't reproduce the bug anymore, so this workaround could be removed.
I guess the bug might have been in the `targetGroupIdx` calculations below.

frontend/src/pages/TaskMapPage.tsx
```diff
@@ -68,47 +69,77 @@ export const TaskMapPage = () => {
...
+      dependencies.map(({ from, to }) => {
+        const targetGroup = measuredRelations.find(({ synergies }) =>
+          synergies.includes(to),
+        );
+
+        return {
+          id: `from-${from}-to-${to}`,
+          source: String(idx),
+          sourceHandle: `from-${from}`,
+          target: String(targetGroup!.id),
```
juhoeerola, line 133:
```suggestion
          target: targetGroup!.id,
```
One extra conversion, that I missed the first time

juhoeerola (approved):
👍 just remove the one conversion before merging

PatricKangasmaki (approved):
Really good job 👍 I tried to break things but this implementation seemed to handle everything really well. Now we just need to add the relation restrictions (for example relation: `1 to 2 to 3 to 1` shouldn't be possible) and then the task map is ready! 🥳 But it's not this PR's thing

One thing I noticed that the lines sometimes cross over each other (where they could be other way around and not cross) but I think it's not worth using possibly multiple working days to fix something like this as we have more important tasks to tackle.

image

IidaKainu:
> I tried to break things but this implementation seemed to handle everything really well

Great to hear! I'll remove the unnecessary conversion Juho found and merge

> and then the task map is ready 🥳

There's also [this significant bug](https://trello.com/c/ffJNPb41/427-bug-dragged-tasks-reset) from rebasing we should fix. But we're getting closer 👍 

> lines sometimes cross over each other

I don't think dagre accounts for positions inside a node/group. So we'd need a custom algorithm and yeah it would take too much time 😄

PR #226, opened 2021-12-23
Add minimum width to table component

juhoeerola:
Given the minimum widths, the buttons will not clip.
The table will scroll sideways when it doesn't fit into the view.

The minimum width is defined as pixels / fr unit. The column widths are
always defined in fr units.


IidaKainu (approved):
Looks good 👍

PR #228, opened 2021-12-29
Add task ratings edition to TaskOverviewPage

IidaKainu:
I diverged from figma's design, as task rating modal displays multiple ratings. There would be no need to find the rating from a (possibly long) rating list if user can edit all of their given ratings through one button.


IidaKainu:
Actually I'll draft this: only admin can currently see the ratings. But feel free to give feedback in the meantime.

PR #229, opened 2021-12-29
Fix redux actions

IidaKainu:


juhoeerola (approved):
Looks good 👍 

Maybe could add explicit `<any>` also on user/actions.ts:107 for consistency
```ts
    return thunkAPI.rejectWithValue(err as AxiosError);
```

PR #230, opened 2021-12-29
Add old invitation cleanup

IidaKainu:
The cleanup is implemented as middleware in post invitations -route

### DIFF
--- a/server/src/api/invitations/invitations.controller.ts
+++ b/server/src/api/invitations/invitations.controller.ts
@@ -8,6 +8,9 @@ import { daysAgo } from '../../utils/date';
 // should FRONTEND_BASE_URL and CORS_ORIGIN be same variable?
 const BASE_URL = process.env.FRONTEND_BASE_URL!;
 
+const deleteOldInvitations = () =>
+  Invitation.query().delete().where('updatedAt', '<', daysAgo(30));
+
 export const getInvitations: RouteHandlerFnc = async (ctx) => {
   ctx.body = await Invitation.query()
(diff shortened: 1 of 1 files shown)


server/src/api/invitations/invitations.controller.ts
```diff
@@ -8,6 +9,14 @@ import { daysAgo } from '../../utils/date';
 // should FRONTEND_BASE_URL and CORS_ORIGIN be same variable?
 const BASE_URL = process.env.FRONTEND_BASE_URL!;
 
+export const deleteOldInvitations = async (
+  _: Context,
+  next: () => Promise<any>,
+) => {
+  await Invitation.query().delete().where('updatedAt', '<', daysAgo(30));
+  await next();
+};
```
juhoeerola, line 18:
```suggestion
const deleteOldInvitations = () =>
  Invitation.query().delete().where('updatedAt', '<', daysAgo(30));
```

I would rather keep this inside the controller, and just add `await deleteOldInvitations();` at the top of `postInvitation()`

juhoeerola (changes requested):

server/src/api/invitations/invitations.controller.ts
IidaKainu: I contemplated whether to use a middleware or not - changed it to the suggested way

juhoeerola (approved):
👍 

I think this way is better here, since the function has nothing inherently to do with the routing or handling the request. It could also be non-exported now.

PR #231, opened 2021-12-30
Update 404 -page

IidaKainu:
Update the page according to the latest figma design

### DIFF
--- a/frontend/src/components/LoginNavBar.tsx
+++ b/frontend/src/components/LoginNavBar.tsx
@@ -1,6 +1,10 @@
 import classNames from 'classnames';
 import { Trans } from 'react-i18next';
+import { shallowEqual, useSelector } from 'react-redux';
 import { Link, useLocation, matchPath } from 'react-router-dom';
+import { RootState } from '../redux/types';
+import { UserInfo } from '../redux/user/types';
+import { userInfoSelector } from '../redux/user/selectors';
 import { paths } from '../routers/paths';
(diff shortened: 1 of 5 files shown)


juhoeerola (changes requested):
Noticed that there is some "state update on unmounted component" warning on the roadmap router component, when coming to the 404 page from `http://localhost:3000/roadmap/1/not-found`. This seems to be from the `useEffect`.

IidaKainu:
> Noticed that there is some "state update on unmounted component" warning on the roadmap router component, when coming to the 404 page from http://localhost:3000/roadmap/1/not-found. This seems to be from the useEffect.

Good catch, it is definitely from roadmap router's useEffect. This (https://github.com/facebook/react/issues/14369#issuecomment-468267798) seems to be the most common fix, yet in our case it has unexpected side-effects. Working on it...

I added the requested changes though

IidaKainu:
Hmm, looks like the "state update on unmounted component" warning [might not be as significant as it sounds in this case](https://github.com/facebook/react/issues/20339), as it [will be removed in the next version of React](https://github.com/facebook/react/pull/22114).
A [comment ](https://github.com/facebook/react/pull/22114#issuecomment-929191048) from the same thread sums up our options: "We’re removing the warning because people work around it with isMounted which defeats the purpose. So we might as well not warn. But the proper solution is cancellation. Or ignoring if you’re lazy about cancelling." I'm personally not sure what to do about it.

juhoeerola (approved):

PR #233, opened 2022-01-04
Update staging branch action to use codedeploy

AleksiKytola:
Deleted old CD script and switch new one from triggering from sandbox pushes to triggering in staging branch pushes


.github/workflows/codedeploy-cd.yml
```diff
@@ -6,7 +6,7 @@ name: AWS Codedeploy CD
 on:
   # Triggers the workflow on push or pull request events but only for the master branch
```
juhoeerola, line 7: I guess the comment has been outdated for a while

juhoeerola (approved):
👍

PR #234, opened 2022-01-05
Show sidebar scrollbar on hover

IidaKainu:
Windows scrollbar would be annoyingly large and would stretch the sidebar on hover pushing other content to the right.

So I propose a custom scrollbar found in figma and showing it on hover. (the scrollbar is used everywhere else too though - stylistically I am not sure if it's an improvement for mac and linux users)

### DIFF
--- a/frontend/src/components/RoadmapSidebar.module.scss
+++ b/frontend/src/components/RoadmapSidebar.module.scss
@@ -8,9 +8,13 @@
 .buttonsContainer {
   @extend .layout-column;
   overflow-y: auto;
+  overflow-x: hidden;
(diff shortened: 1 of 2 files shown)


AleksiKytola (changes requested):
I think the sidebar scrollbar should always show if it doesnt fit on the monitor, instead of only when hovering. This is because the sidebar component itself doesnt look scrollable at all unless the user realizes to hover it.

IidaKainu:
> sidebar scrollbar should always show

Added 👍

AleksiKytola (approved):

PR #235, opened 2022-01-06
Info texts

PatricKangasmaki:
Added missing tooltips. The ones in the settings didn't have specified texts yet so I left them to TBD for now.

### DIFF
--- a/frontend/src/pages/TimeEstimationPage.tsx
+++ b/frontend/src/pages/TimeEstimationPage.tsx
@@ -120,12 +120,12 @@ export const TimeEstimationPage = () => {
           header: t('Estimate milestone durations'),
           content: {
             subHeader: t(
-              'Compare milestone durations with different estimates.',
+              'Compare milestone durations with different estimates',
             ),
             columns: [
               {
                 header: t('Realization'),
-                text: t('system calculated average values over time.'),
+                text: t('system calculated average values over time'),
               },
               {
                 header: t('Comparison'),
@@ -134,7 +134,7 @@ export const TimeEstimationPage = () => {
               {
                 header: t('Custom'),
                 text: t(
-                  'estimate with a custom value for work amount in working days.',
+                  'estimate with a custom value for work amount in working days',
                 ),
               },
             ],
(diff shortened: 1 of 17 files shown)


frontend/src/pages/TimeEstimationPage.tsx
```diff
@@ -110,12 +112,43 @@ export const TimeEstimationPage = () => {
...
+              },
+              {
+                header: t('Custom'),
+                text: t(
+                  'estimate with a custom value for work amount in working days.',
+                ),
+              },
+            ],
+          },
+        },
```
PatricKangasmaki, line 142: I couldn't figure out how to pass react components as params to the modal so I had to do this little workaround where I pass the data as an object with the desired fields and handle it in the modal. This approach should work well with the other infomodals that follow the same design.

IidaKainu (changes requested):
> This approach should work well with the other infomodals that follow the same design.

Since info modals so far follow the same design I agree

Few suggestions
+ I think TimeEstimationPage.module.scss .infoIcon should be changed to correspond with other planner info icons:
```
.infoIcon {
  margin-left: 12px;
}
```

frontend/src/pages/TimeEstimationPage.tsx
juhoeerola: Yeah, the modal payload must be serialisable to json, so components are not possible as params.

The solution here seems good 👍

juhoeerola (changes requested):

PatricKangasmaki:
Really great points, added them all 👍

juhoeerola (approved):

IidaKainu (approved):
👍 

> I think TimeEstimationPage.module.scss .infoIcon should be changed to correspond with other planner info icons

I also added a suggestion for this (not sure why it bothers me so much 😄). Feel free to discard the commit if not needed

PatricKangasmaki:
> I also added a suggestion for this (not sure why it bothers me so much 😄). Feel free to discard the commit if not needed
Ah I forgot to add that part completely! Didn't skip it on purpose, I just went through each comment and added all the suggestions but missed this one while implementing them 😄

PR #236, opened 2022-01-10
Fix dragged task reset -bug

juhoeerola:
The other updates already fetched fresh roadmaps after the update, this one didn't.

The update could be more focused, e.g. only fetching the relations, but this should work fine in the meantime.

### DIFF
--- a/frontend/src/pages/TaskMapPage.tsx
+++ b/frontend/src/pages/TaskMapPage.tsx
@@ -153,8 +153,13 @@ export const TaskMapPage = () => {
     dispatch(roadmapsActions.getRoadmaps());
   };
 
-  const addSynergyRelations = (from: number, to: number[]) =>
-    dispatch(roadmapsActions.addSynergyRelations({ from, to })).unwrap();
+  const addSynergyRelations = async (from: number, to: number[]) => {
+    const ok = await dispatch(
+      roadmapsActions.addSynergyRelations({ from, to }),
+    ).unwrap();
+    if (ok) dispatch(roadmapsActions.getRoadmaps());
+    return ok;
+  };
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/TaskMapPage.tsx
```diff
@@ -153,8 +153,13 @@ export const TaskMapPage = () => {
...
+  const addSynergyRelations = async (from: number, to: number[]) => {
+    const ok = await dispatch(
+      roadmapsActions.addSynergyRelations({ from, to }),
+    ).unwrap();
+    if (ok) dispatch(roadmapsActions.getRoadmaps());
+    return ok;
+  };
 
   const onDragMoveOutside = async (draggedTaskId: number) => {
     const newList = taskRelations
```
PatricKangasmaki, line 165: The original issue seems to be fixed 👍 (_successfully drag a task to a group, navigate elsewhere and back to taskmap_)
But I found out that it also happens when dragging task out of a group. But simply adding `dispatch(roadmapsActions.getRoadmaps());` to the end of the onDragMoveOutside() function should fix that

PatricKangasmaki (changes requested):

frontend/src/pages/TaskMapPage.tsx
juhoeerola: `onConnect()` and `addSynergyRelations()` are the only functions that dispatch api actions to the server, and both refresh the roadmaps at the end. The issue was that the api checked only for 200, when the server returned 204 (no content).

I fixed this for all the api calls that just check the http status.

PatricKangasmaki: Good find 👍

PatricKangasmaki (approved):

PR #238, opened 2022-01-11
Fix awaitsUserRatings for users with no customers represented

AleksiKytola:
Small refactor & comments added as well

### DIFF
--- a/frontend/src/utils/TaskUtils.ts
+++ b/frontend/src/utils/TaskUtils.ts
@@ -237,8 +237,10 @@ export const awaitsUserRatings = <T extends UserInfo | RoadmapUser>(
   roadmap: T extends UserInfo ? number | Roadmap : Roadmap,
 ) => {
   const roadmapId = typeof roadmap === 'number' ? roadmap : roadmap.id;
-  const type = getType(user, roadmapId);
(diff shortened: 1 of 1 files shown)


frontend/src/utils/TaskUtils.ts
```diff
@@ -237,19 +237,28 @@ export const awaitsUserRatings = <T extends UserInfo | RoadmapUser>(
...
     const customers = isUserInfo(user)
       ? user.representativeFor?.filter(
           (customer) => customer.roadmapId === roadmapId,
         )
       : (roadmap as Roadmap).customers?.filter((customer) =>
           customer.representatives?.some((rep) => rep.id === user.id),
         );
+
+    // If user represents no customers, always return false
+    if (!customers || customers.length === 0) return () => false;
```
IidaKainu, line 253: With an empty array [] the 'filter' below should return false anyway, but for readability I think this is fine as it is

IidaKainu (approved):
I can't seem to replicate [the bug](https://trello.com/c/iXsCFOAh/407-business-person-backend-palauttaa-403-kun-antaa-ratingin) on master, but this looks good to me 👍

PR #242, opened 2022-01-13
Center task map on open

juhoeerola:
Fits the map to the view on first load.

Also simplified the position tracking a bit.


IidaKainu (approved):
Works as intended and react flow state handling looks much neater 🔥

PR #244, opened 2022-01-13
Fix user page

juhoeerola:


frontend/src/components/TableTeamMemberRow.tsx
```diff
@@ -72,7 +72,7 @@ export const TableTeamMemberRow: FC<TableRowProps> = ({ member }) => {
 
   return (
     <tr>
-      <td className="styledTd roleIcon">
```
juhoeerola, line 75: This css class is not defined anywhere

IidaKainu (approved):

PR #249, opened 2022-01-20
Fix business users representativeFor bug

IidaKainu:
https://trello.com/c/ayoon1PT/465-business-k%C3%A4ytt%C3%A4j%C3%A4n%C3%A4-yhden-roadmapin-omat-customerit-n%C3%A4kyy-toisessa-roadmapissa

Business users would see their customers from another roadmap in some
cases.

### DIFF
--- a/frontend/src/components/TaskTableUnrated.tsx
+++ b/frontend/src/components/TaskTableUnrated.tsx
@@ -85,11 +85,13 @@ const TableUnratedTaskRow: TableRow<Task> = ({ item: task, style }) => {
     if (type === RoleType.Business) {
       setMissingRatings(
         userInfo?.representativeFor?.filter(
-          (customer) => !ratedByCustomer(customer, userInfo)(task),
+          (customer) =>
+            roadmapId === customer.roadmapId &&
+            !ratedByCustomer(customer, userInfo)(task),
         ),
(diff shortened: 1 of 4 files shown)


frontend/src/components/TaskTableUnrated.tsx
```diff
@@ -85,11 +85,13 @@ const TableUnratedTaskRow: TableRow<Task> = ({ item: task, style }) => {
     if (type === RoleType.Business) {
       setMissingRatings(
         userInfo?.representativeFor?.filter(
-          (customer) => !ratedByCustomer(customer, userInfo)(task),
+          (customer) =>
+            !ratedByCustomer(customer, userInfo)(task) &&
+            roadmapId === customer.roadmapId,
```
juhoeerola, line 90:
```suggestion
            roadmapId === customer.roadmapId &&
            !ratedByCustomer(customer, userInfo)(task),
```

I'd put these checks in the other order, since the other is a simple value comparison

juhoeerola (changes requested):
Good fix 👍 

Just one small suggestion

IidaKainu:
Good catch, added 👍

juhoeerola (approved):

PR #250, opened 2022-01-21
Reset password frontend

PatricKangasmaki:
Some notes:
- Email view is currently accessible in /resetPasswordEmail/:token
- Email view still has things marked as to do since it will most likely need changes when actually implementing it to the email
- All of the pages are using old footers but I figured that creating new footers and applying it to all the other pages as well (email verification etc.) is not in the scope of this PR

### DIFF
--- a/frontend/src/pages/ForgotPasswordPage.tsx
+++ b/frontend/src/pages/ForgotPasswordPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, FormEvent, useMemo } from 'react';
+import { useState, useEffect, FormEvent, useRef } from 'react';
 import { Alert } from 'react-bootstrap';
 import { Trans, useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
@@ -17,78 +17,49 @@ const classes = classNames.bind(css);
 
 export const ForgotPasswordPage = () => {
   const { t } = useTranslation();
-  const defaultDisableTime = 30; // Seconds
+  const defaultDisableSeconds = 30;
 
   const [linkSent, setLinkSent] = useState(false);
   const [email, setEmail] = useState('');
-  const [sendDisabled, setSendDisabled] = useState(false);
-  const [sendDisableTime, setSendDisableTime] = useState(
-    defaultDisableTime / 2,
-  );
   const [resendDisabled, setResendDisabled] = useState(false);
   const [resendDisableTime, setResendDisableTime] = useState(
-    defaultDisableTime,
+    defaultDisableSeconds,
   );
   const [errorMessage, setErrorMessage] = useState('');
   const [isLoading, setIsLoading] = useState(false);
 
-  let timeoutArray = useMemo(() => {
-    const array: NodeJS.Timeout[] = [];
-    return array;
-  }, []);
+  const confirmationError = errorState(useState(''));
 
-  let intervalArray = useMemo(() => {
-    const array: NodeJS.Timeout[] = [];
-    return array;
-  }, []);
+  type Timeout = ReturnType<typeof setTimeout>;
+  const timeout = useRef<Timeout | null>(null);
+  const interval = useRef<Timeout | null>(null);
 
   // Cleanup timers after unmounting
   useEffect(
     () => () => {
-      timeoutArray.forEach((timeout) => {
-        clearTimeout(timeout);
-      });
-      intervalArray.forEach((interval) => {
-        clearInterval(interval);
-      });
+      if (timeout.current !== null) clearTimeout(timeout.current);
+      if (interval.current !== null) clearInterval(interval.current);
     },
-    [timeoutArray, intervalArray],
+    [],
   );
-
-  const confirmationError = {
-    err: errorState(useState('')),
-  };
 
   const sendEmailLink = () => {
     return api.sendPasswordResetLink(email);
   };
 
-  const createDisableTimer = (
-    setDisableTime: React.Dispatch<React.SetStateAction<number>>,
-    setDisabled: React.Dispatch<React.SetStateAction<boolean>>,
-    time: number,
-  ) => {
-    setDisabled(true);
+  const createDisableTimer = () => {
+    setResendDisableTime(defaultDisableSeconds);
+    setResendDisabled(true);
 
     const countdownInterval = setInterval(() => {
-      setDisableTime((previous) => previous - 1);
+      setResendDisableTime((previous) => previous - 1);
     }, 1000);
 
-    const disableTimer = setTimeout(() => {
+    setTimeout(() => {
       clearInterval(countdownInterval);
-
-      // Remove timeout and interval from the arrays after completing them
-      timeoutArray = timeoutArray.splice(timeoutArray.indexOf(disableTimer), 1);
-      intervalArray = intervalArray.splice(
-        intervalArray.indexOf(countdownInterval),
-        1,
-      );
-      setDisabled(false);
-      setDisableTime(time);
-    }, time * 1000);
-
-    timeoutArray.push(disableTimer);
-    intervalArray.push(countdownInterval);
+      setResendDisabled(false);
+      setResendDisableTime(defaultDisableSeconds);
+    }, defaultDisableSeconds * 1000);
   };
 
   const isUniqueError = (err: any) => err.response?.status === 404;
@@ -102,7 +73,7 @@ export const ForgotPasswordPage = () => {
       await sendEmailLink();
     } catch (err) {
       if (isUniqueError(err)) {
-        confirmationError.err.setMessage(t('This email doesn’t exist'));
+        confirmationError.setMessage(t('This email doesn’t exist'));
       } else {
         setErrorMessage(t('Email service error'));
       }
@@ -110,12 +81,7 @@ export const ForgotPasswordPage = () => {
       return;
     }
 
-    setResendDisableTime(defaultDisableTime / 2);
-    createDisableTimer(
-      setSendDisableTime,
-      setSendDisabled,
-      defaultDisableTime / 2,
-    );
+    createDisableTimer();
     setErrorMessage('');
     setIsLoading(false);
     setLinkSent(true);
@@ -123,12 +89,7 @@ export const ForgotPasswordPage = () => {
 
   const handleResend = () => {
     sendEmailLink();
-    setResendDisableTime(defaultDisableTime);
-    createDisableTimer(
-      setResendDisableTime,
-      setResendDisabled,
-      defaultDisableTime,
-    );
+    createDisableTimer();
   };
 
   const sendLinkView = () => (
@@ -150,7 +111,7 @@ export const ForgotPasswordPage = () => {
             type="email"
             placeholder={t('Email')}
             value={email}
-            error={confirmationError.err}
+            error={confirmationError}
             onChange={(e) => setEmail(e.currentTarget.value)}
           />
           <Alert
@@ -167,12 +128,12 @@ export const ForgotPasswordPage = () => {
             <button
               className="button-large"
               type="submit"
-              disabled={sendDisabled}
+              disabled={resendDisabled}
             >
-              {sendDisabled ? (
+              {resendDisabled ? (
                 <Trans
                   i18nKey="Retry timer"
-                  values={{ time: sendDisableTime }}
+                  values={{ time: resendDisableTime }}
                 />
               ) : (
                 <Trans i18nKey="Send link" />
--- a/frontend/src/pages/ForgotPasswordPage.tsx
+++ b/frontend/src/pages/ForgotPasswordPage.tsx
@@ -1,4 +1,5 @@
-import { useState, FormEvent } from 'react';
+import React, { useState, FormEvent } from 'react';
+import { Alert } from 'react-bootstrap';
 import { Trans, useTranslation } from 'react-i18next';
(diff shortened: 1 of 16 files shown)


frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const disableSending = () => {
+    setResendDisabled(true);
+
+    const countdownInterval = setInterval(() => {
+      if (disableTime > 0) setDisableTime((previous) => previous - 1);
+    }, 1000);
+
+    setTimeout(() => setResendDisabled(false), defaultDisableTime * 1000);
+    setTimeout(() => {
+      clearInterval(countdownInterval);
```
juhoeerola, line 38:
```suggestion
    setTimeout(() => {
      setResendDisabled(false);
      clearInterval(countdownInterval);
```

Couldn't these be in the same timeout?

The timeout should also be cancelled when the component is unmounted.

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const defaultDisableTime = 30; // Seconds
+
+  const [view, setView] = useState('Send link');
+  const [email, setEmail] = useState('');
+  const [resendDisabled, setResendDisabled] = useState(false);
+  const [disableTime, setDisableTime] = useState(defaultDisableTime);
+
+  const sendEmailLink = () => {
+    api.sendPasswordResetLink(email);
+  };
```
juhoeerola, line 27: The possible errors should be checked. For example unknown email causes 404, but the page shows as if all went fine.

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const handleResend = () => {
+    sendEmailLink();
+    setDisableTime(defaultDisableTime);
+    disableSending();
+  };
+
+  const sendLinkView = () => {
+    return (
```
juhoeerola, line 57:
```suggestion
  const sendLinkView = () => (
```

This is somewhat a personal preference, but I find the single expression form nicer as it reduces the indentation

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+                  <button
+                    type="button"
+                    className={classes(css.linkButton, css.green)}
+                    onClick={handleResend}
+                  >
+                    <Trans i18nKey="Resend link" />
+                  </button>
+                </>
+              )}
+              {resendDisabled && (
```
juhoeerola, line 132: `resendDisabled ? a : b` would be clearer

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+          </div>
+        </ModalContent>
+      </div>
+    );
+  };
+
+  return (
+    <>
+      {view === 'Send link' && sendLinkView()}
+      {view === 'Link sent' && linkSentView()}
```
juhoeerola, line 148:
```suggestion
      {linkSent ? linkSentView() : sendLinkView()}
```

The `view` would make sense as a boolean (or typed enum e.g. `useState<'a' | 'b' | 'c'>('a')` if there is more than two options)

frontend/src/pages/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,160 @@
...
+          </div>
+        </ModalContent>
+      </div>
+    );
+  };
+
+  return (
+    <>
+      {view === 'Reset password' && resetPasswordView()}
+      {view === 'Done' && doneView()}
```
juhoeerola, line 156:
```suggestion
      {done ? doneView() : resetPasswordView()}
```

frontend/src/pages/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,160 @@
...
+      } else {
+        setView('Done');
+      }
+    });
+  };
+
+  const resetPasswordView = () => {
+    return (
+      <>
+        <div className="formDiv">
```
juhoeerola, line 64:
```suggestion
  const resetPasswordView = () => (
      <div className="formDiv">
```

At least the unnecessary fragment should be removed

frontend/src/pages/ResetPasswordEmail.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Trans } from 'react-i18next';
+import classNames from 'classnames';
+import { ModalContent } from '../components/modals/modalparts/ModalContent';
+import { ModalHeader } from '../components/modals/modalparts/ModalHeader';
+import { ReactComponent as CheckIcon } from '../icons/check_icon.svg';
+import css from './ResetPasswordEmail.module.scss';
+
+const classes = classNames.bind(css);
+
+export const ResetPasswordEmail = () => {
```
juhoeerola, line 11: I don't think we need the "show email in browser" option. Or do we?

Anyway, I would rather have this as a separate pr.

juhoeerola (changes requested):
The flow feels real nice 👌 
One thing I noticed though, is that clicking "send link" -> "change address" -> "send link" -> ... allows skipping the cooldown.


I think the email page should be in separate pr to keep this one a bit more manageable and focused.

PatricKangasmaki:
Great points 👍 Added all your suggestions

> One thing I noticed though, is that clicking "send link" -> "change address" -> "send link" -> ... allows skipping the cooldown.

Added separate cooldown for the send link button on the first page. Made it 15 sec so it prevents spamming but shouldn't affect real cases where you accidentally typed wrong address too much.

> I think the email page should be in separate pr to keep this one a bit more manageable and focused.  

Removed the email page from this pr

> The timeout should also be cancelled when the component is unmounted.  

Missed this part in the previous fixes, it's under progress now.

frontend/src/pages/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,153 @@
...
+          setErrorMessage(t('Token expired'));
+        else if (res.payload?.response?.status === 404)
+          setErrorMessage(t('Token not found'));
+        else if (res.payload) {
+          setErrorMessage(t('Error message', { error: res.payload.message }));
+        }
+      } else {
+        setDone(true);
+      }
+    });
```
IidaKainu, line 58: The nesting here could be reduced (generally makes code easier to understand). Inverting an "if" statement is a great way achieve it. [Example](https://www.refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html)
```suggestion
    const res = await dispatch(userActions.resetPassword({ token, password }));
    if (!userActions.resetPassword.rejected.match(res)) {
      setDone(true);
      return;
    }
    if (res.payload?.response?.status === 403)
      setErrorMessage(t('Token expired'));
    else if (res.payload?.response?.status === 404)
      setErrorMessage(t('Token not found'));
    else if (res.payload)
      setErrorMessage(t('Error message', { error: res.payload.message }));
```

IidaKainu (changes requested):
I have to agree the user flow is great 🔥

frontend/src/pages/ResetPasswordPage.tsx
PatricKangasmaki: Good point, your version looks a lot cleaner when there's no 3 nested if's 👍 Added the suggestion

IidaKainu (approved):
Approved on my behalf 👍

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,251 @@
...
+        confirmationError.err.setMessage(t('This email doesn’t exist'));
+      } else {
+        setErrorMessage(t('Email service error'));
+      }
+      setIsLoading(false);
+      return;
+    }
+
+    setResendDisableTime(defaultDisableTime / 2);
+    createDisableTimer(
```
juhoeerola, line 114:
```suggestion
    createDisableTimer(
```

`createDisableTime()` could set the resend disable time, since the `time` parameter is just that

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const [linkSent, setLinkSent] = useState(false);
+  const [email, setEmail] = useState('');
+  const [sendDisabled, setSendDisabled] = useState(false);
+  const [sendDisableTime, setSendDisableTime] = useState(
+    defaultDisableTime / 2,
+  );
+  const [resendDisabled, setResendDisabled] = useState(false);
+  const [resendDisableTime, setResendDisableTime] = useState(
+    defaultDisableTime,
+  );
```
juhoeerola, line 31: I think the two flags & times should use just one and same flag & time. Just setting the time to different initial value from different places.

This would simplify the code and keep the timer consistent when clicking to "change address" after resending.

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,251 @@
...
+import { LoadingSpinner } from '../components/LoadingSpinner';
+import { ReactComponent as MailIcon } from '../icons/mail_icon.svg';
+import { api } from '../api/api';
+import css from './ForgotPassword.module.scss';
+
+const classes = classNames.bind(css);
+
+export const ForgotPasswordPage = () => {
+  const { t } = useTranslation();
+  const defaultDisableTime = 30; // Seconds
```
juhoeerola, line 20:
```suggestion
  const defaultDisableSeconds = 30;
```

The comment could be moved to the name, then it would also be visible in each use

juhoeerola (changes requested):
Mostly about the timeout handling, and couple other minor suggestions.

For the other parts looks good to me.

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,251 @@
...
+    () => () => {
+      timeoutArray.forEach((timeout) => {
+        clearTimeout(timeout);
+      });
+      intervalArray.forEach((interval) => {
+        clearInterval(interval);
+      });
+    },
+    [timeoutArray, intervalArray],
+  );
```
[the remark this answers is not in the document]
PatricKangasmaki: Do you mean generally or in this specific scenario? My reasoning for the two different timers here is that I made them to serve different purposes.

The first timeout is for setting timer to the "Send link" -button on the sendLinkView. It starts when user requests a link on that first view. The reason for this timeout is that you can't spam emails by sending link -> change address -> send link etc. If user is using this page correctly, they should never even know about this timeout, that's why it's 15 seconds instead of the 30 which is on the second view as 30 seconds could have effect on normal use as well eg. accidentally entering wrong email and realizing it in 10 seconds and then having to wait another 20. This is just a small protection against malicious use.

The second timeout is for the "Resend link" -button on the linkSentView. It starts when users requests a new link (after the first email didn't supposedly arrive). The reason for this being different timer than the sendLink one is that it didn't look as good and smooth (to me) if the "Didn’t receive the email? Resend link" was already disabled when entering the page the first time.  

And that's why both of these timers could be active at the same time since they get set from different actions and I had to clean them both in the cleanup

frontend/src/pages/ForgotPasswordPage.tsx
PatricKangasmaki: This continues from the above explanation. The sendDisableTime can be going 15->14->13->12 etc. while the resendDisableTime is going from 30->29->... so they needed their own times in this particular implementation (at least I think they are needed).

frontend/src/pages/ForgotPasswordPage.tsx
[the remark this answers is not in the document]
PatricKangasmaki: Maybe I'm thinking about these timers way too complicatedly and the pages could just share the timers 😬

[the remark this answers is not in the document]
juhoeerola: > the pages could just share the timers

This is what I was thinking. It would also keep the time consistent when the user navigates to change email during the 30 s timer.

> This is just a small protection against malicious use.

Any enforcement of the send rate must be done on the server side anyway, but it's good that the frontend indicates that the user should not spam the button if they didn't receive the email in a heartbeat.

PatricKangasmaki:
Added all the changes 👍 There's now just one timer which controls both send and resend buttons

frontend/src/pages/ForgotPasswordPage.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+    createDisableTimer();
+    setErrorMessage('');
+    setIsLoading(false);
+    setLinkSent(true);
+  };
+
+  const handleResend = () => {
+    sendEmailLink();
+    createDisableTimer();
```
juhoeerola, line 92:
```suggestion
    createDisableTimer(defaultDisableSeconds / 2);
    setErrorMessage('');
    setIsLoading(false);
    setLinkSent(true);
  };

  const handleResend = () => {
    sendEmailLink();
    createDisableTimer(defaultDisableSeconds);
```

The time could still be a parameter, so typo in email isn't quite as long wait

juhoeerola (approved):
Looks good 👍 

The time could be parametrised, but either way is fine, your call

PatricKangasmaki:
> The time could be parametrised, but either way is fine, your call

Added this 👍

PR #251, opened 2022-01-21
Leave roadmap button

AleksiKytola:


server/src/api/roadmaps/roadmaps.controller.ts
```diff
@@ -96,3 +96,12 @@ export const deleteRoadmaps: RouteHandlerFnc = async (ctx) => {
 
   ctx.status = numDeleted == 1 ? 200 : 404;
 };
+
+export const leaveRoadmap: RouteHandlerFnc = async (ctx) => {
+  const numDeleted = await Role.query().deleteById([
+    Number(ctx.state.user!.id),
```
IidaKainu, line 102: I'd add undefined check that's used in many other routes as well
```suggestion
  if (!ctx.state.user) throw new Error('User is required');
  const numDeleted = await Role.query().deleteById([
    Number(ctx.state.user.id),
```

IidaKainu (changes requested):
Found a few suggestions, otherwise looks good to me 👍

frontend/src/colors.module.scss
```diff
@@ -22,6 +23,7 @@ $COLOR_BLACK5: #f5f5f5;
   cloud: $COLOR_CLOUD;
   raspberry: $COLOR_RASPBERRY;
   mandy: $COLOR_MANDY;
+  mandy_hover: $COLOR_MANDY;
```
juhoeerola, line 26:
```suggestion
  mandy_hover: $COLOR_MANDY_HOVER;
```

frontend/src/components/UserInfoCard.tsx
```diff
@@ -18,13 +18,31 @@ import { ModalTypes } from './modals/types';
...
 const classes = classNames.bind(css);
 
 const ProjectRow: TableRow<RoadmapRole> = ({ item: roadmapRole, style }) => {
   const { t } = useTranslation();
+  const dispatch = useDispatch();
   const { roadmapId, type } = roadmapRole;
   const roadmap = useSelector(idRoadmapSelector(roadmapId));
+
+  const leaveProject = useCallback(
+    (event) => {
```
juhoeerola, line 32:
```suggestion
    (event: SyntheticEvent) => {
```

For some reason eslint doesn't complain about "implicit any" when wrapped in `useCallback`

juhoeerola (changes requested):
The last admin should probably not be able to leave the roadmap. This would otherwise leave ownerless roadmaps floating around.

I think the options for last admin would be:
- delete the whole roadmap
- assign another team member as admin, and then leave

frontend/src/components/UserInfoCard.module.scss
```diff
@@ -37,7 +37,7 @@
 
 .deleteButton {
   @extend .button-small-filled;
-  background: $COLOR_RASPBERRY;
+  background: $COLOR_MANDY;
   &:hover:not(.cancel, :disabled),
   &:focus:not(.cancel, :disabled) {
     background: $COLOR_MANDY;
```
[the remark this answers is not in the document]
AleksiKytola: A good catch, fixed

AleksiKytola:
Applied stylistic fixes, blocked leaving roadmaps if the current user is the last admin. Rebased.

server/src/api/roadmaps/roadmaps.controller.ts
```diff
@@ -96,3 +96,34 @@ export const deleteRoadmaps: RouteHandlerFnc = async (ctx) => {
...
   ctx.status = numDeleted == 1 ? 200 : 404;
 };
+
+export const leaveRoadmap: RouteHandlerFnc = async (ctx) => {
+  if (!ctx.state.user) throw new Error('User is required');
+  await Role.transaction(async (trx) => {
+    const userRole = await Role.query(trx)
+      .where('userId', ctx.state.user!.id)
+      .andWhere('roadmapId', ctx.params.roadmapId)
+      .first();
```
juhoeerola, line 106:
```suggestion
      .findById([ctx.state.user!.id, Number(ctx.params.roadmapId)]);
```

Should also check if the role was found

server/src/api/roadmaps/roadmaps.controller.ts
```diff
@@ -96,3 +96,34 @@ export const deleteRoadmaps: RouteHandlerFnc = async (ctx) => {
...
+        ctx.body =
+          'Cannot leave as the last admin of roadmap, delete the whole roadmap instead.';
+        return;
+      }
+    }
+
+    const numDeleted = await Role.query(trx).deleteById([
+      Number(ctx.state.user!.id),
+      Number(ctx.params.roadmapId),
+    ]);
```
juhoeerola, line 125:
```suggestion
    const numDeleted = await role.$query(trx).delete();
```

This would avoid duplicating the query criteria from above.

server/src/api/roadmaps/roadmaps.controller.ts
```diff
@@ -96,3 +96,34 @@ export const deleteRoadmaps: RouteHandlerFnc = async (ctx) => {
...
+
+    ctx.status = numDeleted == 1 ? 200 : 404;
```
juhoeerola, line 127:
```suggestion
    ctx.status = numDeleted == 1 ? 200 : 500;
```

This should always be 1, so otherwise a server error might be more appropriate

juhoeerola (changes requested):
The frontend looks fine, couple fixes/improvements for the server.

The test failure seems to be incorrect assertion.

juhoeerola (approved):

PR #253, opened 2022-01-24
Add delete roadmap button to settings page

AleksiKytola:


frontend/src/pages/ConfigurationPage.tsx
```diff
@@ -142,6 +153,21 @@ const RoadmapConfigurationPageComponent = ({
           </button>
         </span>
       </div>
+      {getType(userInfo, currentRoadmap?.id) === RoleType.Admin && (
```
IidaKainu, line 156: The getType result should be stored in a variable, as it is currently computed twice in the component.
```suggestion
      {type === RoleType.Admin && (
```

IidaKainu (changes requested):
Just a small request, otherwise looks good 👍

IidaKainu (approved):

PR #254, opened 2022-01-24
Remove order dependency from tests

juhoeerola:
Previously some tests depended on consistent order from database queries
and api responses, where no ordering was required for the tested functionality.

### DIFF
--- a/server/tests/customer.test.ts
+++ b/server/tests/customer.test.ts
@@ -13,6 +13,9 @@ import {
 } from './testUtils';
 chai.use(chaiHttp);
 
+const byId = <T extends { id: number } | { id: string }>(a: T, b: T) =>
+  a.id.toString().localeCompare(b.id.toString());
+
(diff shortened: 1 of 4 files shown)


server/tests/customer.test.ts
```diff
@@ -13,6 +13,9 @@ import {
 } from './testUtils';
 chai.use(chaiHttp);
 
+const byId = <T extends { id: number } | { id: string }>(a: T, b: T) =>
+  a.toString().localeCompare(b.toString());
```
IidaKainu, line 17: I may not be interpreting this correctly, but shouldn't this be `a.id.toString().localeCompare(b.id.toString());` if we are comparing ids?

juhoeerola: yes it should, good catch 👍

IidaKainu (approved):
Great fix 👍

PR #256, opened 2022-01-26
Task map validation

juhoeerola:
While drawing a new connection:
- tasks that can not be connected to are grayed out
- handles that can accept the connection are highlighted

While dragging task from synergy group to another, unavailable groups are grayed out.

On the server side, the graph is validated each time new relations are added.
Deleting relations will never invalidate the graph.

### DIFF
--- a/frontend/src/components/TaskMapTask.module.scss
+++ b/frontend/src/components/TaskMapTask.module.scss
@@ -13,6 +13,10 @@
   position: relative;
   align-items: center;
 
+  &.unavailable {
+    filter: grayscale(100%) opacity(20%);
+  }
+
   button {
     visibility: hidden;
     padding: 0;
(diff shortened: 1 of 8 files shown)


AleksiKytola (approved):
Looked good in the demo, I'm not particularly familiar with the task map but on a quick glance the code looks good

PatricKangasmaki (changes requested):
You can currently drag relations to tasks that are in the same group (Even though the highlight says you can't)

image


image

juhoeerola:
> You can currently drag relations to tasks that are in the same group

The backend correctly rejects this for me, so I couldn't reproduce this.

But sure the frontend shouldn't even try sending that request, I'll fix that.

PatricKangasmaki:
> The backend correctly rejects this for me

Ah, I probably had old server up from some other branch

juhoeerola:
Should be fixed now.

The state was clearing between the connection ending and validation, which would suggest that there are some unnecessary rendering happening. But that should no longer affect the correctness.

PatricKangasmaki (approved):

PR #258, opened 2022-01-26
Sort roadmaps in ProjectOverviewPage

IidaKainu:
Sort by alphabetical order of roadmap names

### DIFF
--- a/frontend/src/pages/ProjectOverviewPage.tsx
+++ b/frontend/src/pages/ProjectOverviewPage.tsx
@@ -22,6 +22,9 @@ export const ProjectOverviewComponent = () => {
     roadmapsSelector,
     shallowEqual,
   );
+  const sortedRoadmaps = [...(roadmaps || [])].sort((a, b) =>
+    a.name.localeCompare(b.name),
+  );
 
   useEffect(() => {
     if (!roadmaps) dispatch(roadmapsActions.getRoadmaps());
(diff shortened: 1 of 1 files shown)


frontend/src/pages/ProjectOverviewPage.tsx
```diff
@@ -22,11 +22,21 @@ export const ProjectOverviewComponent = () => {
...
     if (!roadmaps) dispatch(roadmapsActions.getRoadmaps());
   }, [dispatch, roadmaps]);
 
+  useEffect(() => {
+    if (roadmaps)
+      setSortedRoadmaps(
+        [...roadmaps].sort((a, b) => a.name.localeCompare(b.name)),
+      );
+  }, [roadmaps]);
+
```
AleksiKytola, line 39: Im not sure if useState and useEffect are even needed here, probably the only thing re-rendering the component would be the roadmaps changing and the selector triggering. But either way is fine with me
```
const roadmaps = useSelector<RootState, Roadmap[] | undefined>(
    roadmapsSelector,
    shallowEqual,
  );
const sortedRoadmaps = [...roadmaps].sort((a, b) => a.name.localeCompare(b.name)),
```

AleksiKytola (approved):
Im not sure if the `

IidaKainu:
Yeah the useEffect is likely not needed, removed it

PR #262, opened 2022-01-27
Fix member invitation bug in project creation

PatricKangasmaki:


AleksiKytola (approved):

PatricKangasmaki:
Adjusted this fix as master branch had some changes which modified sendInvitation action. Does exactly the same as before

PR #264, opened 2022-01-27
Fix bug when using multiple tabs and switching roadmaps

AleksiKytola:
I couldn't fix the edge case of another tab not navigating from /overview to /roadmap/id/dashboard if the state gets synced. Im guessing its because of checking for an existing `previousRoadmapId` before running the useEffect, but since `undefined` is a valid roadmapId (not selected) the useEffect/usePrevious cant tell the difference between the first render & intentionally null/undefined roadmap
`if (previousRoadmapId && selectedRoadmapId !== previousRoadmapId) {`

removing the check causes infinite loops and exploding galaxies. It might be fixable if we refactor our state to use `null` for intentionally nonexistent values to tell them apart from `undefined`

### DIFF
--- a/frontend/src/routers/RoadmapRouter.tsx
+++ b/frontend/src/routers/RoadmapRouter.tsx
@@ -4,6 +4,7 @@ import {
   Redirect,
   Route,
   Switch,
+  useHistory,
   useParams,
   useRouteMatch,
 } from 'react-router-dom';
@@ -22,6 +23,7 @@ import {
   chosenRoadmapSelector,
   roadmapUsersSelector,
   allCustomersSelector,
+  chosenRoadmapIdSelector,
 } from '../redux/roadmaps/selectors';
 import { RoadmapUser, Customer, Roadmap } from '../redux/roadmaps/types';
 import { RootState } from '../redux/types';
@@ -30,6 +32,7 @@ import { paths } from './paths';
 import { PlannerPageRouter } from './PlannerPageRouter';
 import { TasksPageRouter } from './TasksPageRouter';
 import '../shared.scss';
+import { usePrevious } from '../utils/usePrevious';
 
 const routes = [
   {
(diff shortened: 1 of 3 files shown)


frontend/src/routers/RoadmapRouter.tsx
```diff
@@ -91,9 +99,42 @@ const RoadmapRouterComponent = ({ userInfo }: { userInfo: UserInfo }) => {
...
+      dispatch(roadmapsActions.selectCurrentRoadmap(+urlRoadmapId));
     }
+  }, [dispatch, urlRoadmapId]);
+
+  useEffect(() => {
+    // If selectedRoadmapId changes, update the url to match that
+    // (for cases where another browser tab switches roadmapId and the state gets synced into this tab)
+    if (previousRoadmapId && selectedRoadmapId !== previousRoadmapId) {
+      let newPath: string;
+      const roadmapIdRegex = new RegExp('roadmap/\\d+/');
```
juhoeerola, line 112:
```suggestion
      const roadmapIdRegex = new RegExp('^/roadmap/\\d+/');
```

The regex could be anchored, even though we are unlikely to add other routes that would match

juhoeerola (commented):
Seems to work well, and I'd say the overview case is acceptable.

Couple suggestions

frontend/src/routers/RoadmapRouter.tsx
juhoeerola: The path has a leading `/`

frontend/src/routers/RoadmapRouter.tsx
```diff
@@ -91,9 +99,41 @@ const RoadmapRouterComponent = ({ userInfo }: { userInfo: UserInfo }) => {
...
+    if (!previousRoadmapId || selectedRoadmapId === previousRoadmapId) return;
+    if (selectedRoadmapId === undefined) {
+      newPath = '/overview';
+    } else if (
+      selectedRoadmapId !== previousRoadmapId &&
+      history.location.pathname.match(roadmapIdRegex)
+    ) {
+      newPath = history.location.pathname.replace(
+        roadmapIdRegex,
+        `roadmap/${selectedRoadmapId}/`,
```
juhoeerola, line 121:
```suggestion
        `/roadmap/${selectedRoadmapId}/`,
```

Anchoring the regex requires the leading `/` here also

frontend/src/routers/RoadmapRouter.tsx
```diff
@@ -91,9 +99,41 @@ const RoadmapRouterComponent = ({ userInfo }: { userInfo: UserInfo }) => {
...
+  useEffect(() => {
+    // If selectedRoadmapId changes, update the url to match that
+    // (for cases where another browser tab switches roadmapId and the state gets synced into this tab)
+    let newPath: string;
+    const roadmapIdRegex = new RegExp('^roadmap/\\d+/');
+    if (!previousRoadmapId || selectedRoadmapId === previousRoadmapId) return;
+    if (selectedRoadmapId === undefined) {
+      newPath = '/overview';
+    } else if (
+      selectedRoadmapId !== previousRoadmapId &&
```
juhoeerola, line 116:
```suggestion
```

This is already checked couple lines above in the guard clause

frontend/src/routers/RoadmapRouter.tsx
AleksiKytola: Weird, it somehow still worked in my testing, I'll change it

PR #265, opened 2022-01-28
Rewording work -> complexity

juhoeerola:
"Work" term is now only used in connection to time estimation (workdays being the time unit), otherwise tasks are rated etc. on "complexity"

### DIFF
--- a/frontend/src/components/TaskHeatmap.module.scss
+++ b/frontend/src/components/TaskHeatmap.module.scss
@@ -42,7 +42,7 @@
   .xaxislabel {
     @extend .typography-pre-title;
     position: relative;
-    left: 68px;
+    left: 32px;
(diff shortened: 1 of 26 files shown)


frontend/src/components/TaskHeatmap.module.scss
```diff
@@ -42,7 +42,7 @@
   .xaxislabel {
     @extend .typography-pre-title;
     position: relative;
-    left: 68px;
+    left: 32px;
```
juhoeerola, line 45: This is very brittle at the moment, the label should be actually centred in the available space.

AleksiKytola (approved):
I think the commits that contain work -> complexity rewording could be squashed together, other than that it looks good :+1:

PR #267, opened 2022-01-31
Add RepresentativeTable to client overview

IidaKainu:


frontend/src/pages/ClientOverviewPage.tsx
```diff
@@ -115,6 +124,11 @@ const ClientOverview: FC<{
         metrics={metrics}
         data={clientData}
       />
+      {representatives.length && (
```
juhoeerola, line 127:
```suggestion
      {representatives.length > 0 && (
```

This would display literal "0"

frontend/src/components/RepresentativeTable.module.scss
```diff
@@ -0,0 +1,5 @@
+@import '../shared.scss';
+
+.memberIcon path {
+  fill: $COLOR_BLACK40;
+}
```
juhoeerola, line 5: This is now repeated in quite a few places, maybe it would be time to make a component for the member icon

frontend/src/components/RepresentativeTable.tsx
```diff
@@ -0,0 +1,39 @@
...
+import css from './RepresentativeTable.module.scss';
+
+const classes = classNames.bind(css);
+
+const RepresentativeRow: TableRow<RoadmapUser> = ({ item: user, style }) => (
+  <div style={style} className={classes(css.virtualizedTableRow)}>
+    <div className={classes(css.memberIcon)}>
+      {user.type === RoleType.Admin && <StarSharpIcon />}
+      {user.type === RoleType.Business && <BusinessIcon />}
+    </div>
```
juhoeerola, line 18: This could use a component

Then it would be simple to add e.g. hover that shows the role in text

juhoeerola (changes requested):
Looking good 👍 

I'd say this can be merged after the zero fix.

The refactor would be fine here or separately.

frontend/src/pages/ClientOverviewPage.tsx
IidaKainu: Definitely made this mistake before too, thanks for catching it

IidaKainu:
Added fixes

> maybe it would be time to make a component for the member icon

I agree: added a card for this. I'll do it in another pr

> Then it would be simple to add e.g. hover that shows the role in text

Tooltips for the icons are a great idea 👍

juhoeerola (approved):

PR #268, opened 2022-02-01
Roletype icon component

IidaKainu:
The component displays an icon corresponding to the RoleType.
- Tooltip versions of the RoleIcon's are added to places where the role name cannot be deduced from nearby elements.
  - The tooltips are a bit out of align in lists made with the table-component, such as representatives and invitations lists. Does it look bad a.k.a should I spend time fixing this?

### DIFF
--- a/frontend/src/components/NavBar.module.scss
+++ b/frontend/src/components/NavBar.module.scss
@@ -86,11 +86,3 @@
   margin-right: 10px;
   gap: 2px;
 }
-
-.memberIcon path {
-  fill: $COLOR_BLACK100;
-}
-
-.memberIcon svg {
-  width: 18px;
-}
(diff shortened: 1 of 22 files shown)


PatricKangasmaki (changes requested):
> The tooltips are a bit out of align in lists made with the table-component

I think this would be good to fix. Especially in the invitations list as the tooltips move depending on the screen width (looks like the position is tied to the width of the role-column). 
image


I think email notification modal could display role icons as well. There might be situations where someone wants to notify all the developers but no one else. Otherwise this is looking great 👍

IidaKainu:
> I think this would be good to fix.

Looks like I only had to wrap the icon in a flex-container. Solved.

> I think email notification modal could display role icons as well. There might be situations where someone wants to notify all the developers but no one else.

Yeah it could be useful in some cases. 👍 Added. I made some changes to the notification modal useEffect, since customer's representatives do not include the type.

PatricKangasmaki (approved):
👍

PR #269, opened 2022-02-01
RTK-query

juhoeerola:
Testing if using RTK-query would be feasible refactor. Conclusion: it would be.

I think this is at a good state to consider now. The two remaining items in the todo are relatively independent changes. I'd leave those as separate tasks, and make sure the current implementation is working fine before going further with those.

The user routes require some additional care, as those handle the auth state, and I'm not even quite sure if it is worth moving to use rtk-query.

todo:
- [x] add permission checks back to avoid unnecessary queries
- [x] add null safety back (don't query with `undefined` roadmap id)
- [ ] refactor user routes
- [x] use selectFromQuery to select subset of query (e.g. single roadmap)
- [ ] maybe refactor some local copies to use optimistic updates instead

### DIFF
--- a/frontend/src/utils/TaskRelationUtils.ts
+++ b/frontend/src/utils/TaskRelationUtils.ts
@@ -123,7 +123,7 @@ export const blockedGroups = (task: number, graph: GroupedRelation[]) => {
       const direction = from === task ? 'target' : 'source';
       reachable(start, graph, direction).forEach((x) => {
         const node = taskToGroup.get(x);
-        if (node) visited.add(node.toString());
+        if (node !== undefined) visited.add(node.toString());
       });
     });
   });


frontend/src/components/RatingTableComplexity.tsx
```diff
@@ -23,8 +23,11 @@ const TableComplexityRatingRow: RatingRow = ({
   onEdit,
 }) => {
   const roadmapId = useSelector(chosenRoadmapIdSelector);
-  const { data } = apiV2.useGetRoadmapUsersQuery(roadmapId ?? skipToken);
-  const user = data?.find(({ id }) => id === rating.createdByUser);
+  const { user } = apiV2.useGetRoadmapUsersQuery(roadmapId ?? skipToken, {
+    selectFromResult: ({ data }) => ({
+      user: data?.find(({ id }) => id === rating.createdByUser),
+    }),
```
AleksiKytola, line 29: Mietin että pitäisikö tällaiset yleisesti käytetyt "selectorit" koota jonnekkin yhteiseen moduliin

AleksiKytola (commented):
Looking good so far. I can't really find anything to add other than perhaps separating the often used 'selectors' for `selectFromResult` into a module.

frontend/src/components/RatingTableComplexity.tsx
juhoeerola: Lisäsin id:lle kun se oli oikeastaan ainoa yleisemmin käytetty selektori, toi ainoastaan muuttaa `data` kentän ja välittää muut sellaisenaan

juhoeerola:
did rebase & merge locally

PR #270, opened 2022-02-03
Task overview relations

IidaKainu:


IidaKainu:
Thanks for spotting the relation filtering bug: fixed

> The columns quite easily go to the second row (~ 1460 px wide window), this is especially noticeable when there is no "precedes" relations and the content wraps seemingly unnecessarily. The content may also jump between 1 or 2 rows depending on how many columns have relations, even though the widths are (somewhat) fixed.

How about having no wrapping with the overview lists? I'd like to keep the list widths fixed, since it's easier to scan relations while browsing tasks. I pushed a test commit with this implementation: how does it look?

PR #271, opened 2022-02-08
Forgot password email

PatricKangasmaki:
`frontend/src/components/emails/ResetPasswordEmail.tsx` creates the email html and console logs it and I manually added it to `server/src/utils/emailMessages.ts`. ResetPasswordEmail.tsx serves no other purpose as it can't be accessed from anywhere (I had routing to it while developing it to test it and to get the console log though).

CSS has to be inline and it looks kinda raw as emails don't accept most of the css properties.

@juhoeerola do you know if this email format (`<div> stuff here </div`) is enough for amazon ses or should it also have the html and doctype etc. tags?


juhoeerola:
> is enough for amazon ses or should it also have the html and doctype etc. tags?

afaik. ses doesn't care about the html content, it will just set the correct headers etc. when sending.
So, the html content should be whatever email clients expect. I'd make it complete html document e.i. include doctype and `<html><head></head><body>...` at least.

frontend/src/components/emails/ResetPasswordEmail.tsx
```diff
@@ -0,0 +1,71 @@
...
+      </div>
+      <div style={{ ...textStyle, marginTop: '40px', marginBottom: '-40px' }}>
+        <i>
+          Didn’t order the password reset link? You can ignore this message.
+        </i>
+      </div>
+    </EmailTemplate>
+  );
+  const html = ReactDOMServer.renderToStaticMarkup(page());
+  console.log(html.toString());
```
IidaKainu, line 69: Just letting you know here is an extra console.log

juhoeerola (approved):
Looks good to me 👍 

The process *could* be automated bit more, but I doubt that would benefit us anytime soon. Maybe next time we need to add new emails...

PR #272, opened 2022-02-08
Fix constraint on task table

juhoeerola:
This adds the roadmapId to the constraint, so that the same task can be
imported to multiple roadmaps.


PatricKangasmaki (approved):
👍

PR #273, opened 2022-02-09
Add migration to rescale ratings from 0-10 to 0-5

AleksiKytola:


server/src/migrations/20220209093644_rescaleRatings.ts
```diff
@@ -0,0 +1,9 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  knex('taskratings').update({ value: knex.raw('taskratings.value / 2') });
```
juhoeerola, line 4: The scale of actual given ratings is [1, 5] (previously [1, 10]), so this would map some ratings to 0.

(value - 1) / 2 + 1 should map it correctly (1, 2) -> 1, (3, 4) -> 2, ..., (9, 10) -> 5

The other way works fine (biasing towards higher ratings), but obviously some resolution is lost there

juhoeerola (changes requested):
With the mapping fix, I'd say this is good to go 👍

juhoeerola (approved):

PR #281, opened 2022-02-16
Task map list for individual tasks

juhoeerola:
This splits the tasks in task map to staged (having relations or manually added), and unstaged (all the rest) tasks.
The unstaged tasks in a list for now, but that should be simple to change.

The unstaged tasks can be staged by dragging to the task map, either individually or into existing group.

### DIFF
--- a/frontend/src/components/SortableTaskList.tsx
+++ b/frontend/src/components/SortableTaskList.tsx
@@ -11,11 +11,23 @@ export const SortableTaskList: FC<{
   listId: string;
   tasks: Task[];
   disableDragging: boolean;
+  isDropDisabled?: boolean;
   showRatings?: boolean;
   className?: string;
-}> = ({ listId, tasks, disableDragging, showRatings, className }) => (
+}> = ({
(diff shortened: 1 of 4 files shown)


IidaKainu (approved):
🔥 (+ got how the useEffects work now)

No need to do this now, but in the future we could:
1. Try to virtualize the task list
2. Add functionality for moving tasks to unstaged list
3. Move some functions to TaskRelationUtils from TaskMapPage (the long file length is starting to affect the readability)

What do you think?

juhoeerola:
All good points that should be considered.

I now added 2., as we talked that this functionality would be good to have.

I did try 1., but didn't continue further when it caused problems with the drag and drop.
1. and 3. would be good to do, but I'd say better separately from this PR.

IidaKainu:
> I now added 2., as we talked that this functionality would be good to have.

Great, this works well! I made Trello cards about the other tasks.

IidaKainu (approved):

PR #282, opened 2022-02-16
Add consistent order to customer weights list

juhoeerola:
The rtk query made the customers jump around on updates.
This adds a consistent order to the list.


IidaKainu (approved):
👍

PR #284, opened 2022-02-18
SocketIo realtime updates

AleksiKytola:
To test locally, open two separate browsers, such as Firefox and Chrome. Login as two separate accounts such as admin.person1 and admin.person2 and view the realtime updates happen. Lots of console logs are still included in frontend and backend


frontend/src/api/api.ts
```diff
@@ -401,6 +400,30 @@ export const apiV2 = createApi({
         data: { from, to },
       }),
     }),
+    refetchTasks: build.mutation<null, null>({
```
juhoeerola, line 403:
```suggestion
    refetchTasks: build.mutation<null, void>({
```

The parameter could be void, then there is no need to pass `null` for each call.

juhoeerola (changes requested):
Couple minor suggestions, but overall looks good

Also, the console logging should be toned down.

PR #286, opened 2022-02-19
Add customer dots to notification modal

PatricKangasmaki:


frontend/src/components/modals/NotifyUsersModal.module.scss
```diff
@@ -17,10 +17,22 @@ hr {
...
+  flex-wrap: wrap;
+  max-width: 135px;
+  margin-left: 10px;
+  margin-bottom: 15px;
+  gap: 3px;
+}
+
+.checkboxDiv {
+  margin-top: -11px;
+}
```
PatricKangasmaki, line 38: This is to align the checkbox and email with the customer dots when there are multiple rows of customers

PatricKangasmaki:
Last push fixed console warning: Each child in a list should have a unique "key" prop.

IidaKainu (commented):
Just a small change request, otherwise looks good 👍

juhoeerola:
Seems like this only has one truly new commit (compared to master), so I'd suggest rebasing and dropping those duplicates from this pr before merging. That should also resolve the conflicts.

PatricKangasmaki:
Added both suggestions 👍

juhoeerola (approved):

PR #288, opened 2022-02-21
Rework milestone editor logic, fix issue with rtk-query

AleksiKytola:
Removed unnecessary optimistic update logic from milestone editor.
Made backend patchVersions route remove tasks from any previous versions when they are added to a new version.
Cleaned up milestone editor and fixed issues with rtk-query

### DIFF
--- a/frontend/src/pages/MilestonesEditor.tsx
+++ b/frontend/src/pages/MilestonesEditor.tsx
@@ -55,45 +55,40 @@ export const MilestonesEditor = () => {
   const { t } = useTranslation();
   const roadmapId = useSelector(chosenRoadmapIdSelector);
   const { data: tasks } = apiV2.useGetTasksQuery(roadmapId ?? skipToken);
-  const [patchVersion] = apiV2.usePatchVersionMutation();
+  const [
+    patchVersion,
+    { isLoading: disableDrag, isError },
+  ] = apiV2.usePatchVersionMutation();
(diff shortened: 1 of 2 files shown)


frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -167,52 +160,39 @@ export const MilestonesEditor = () => {
...
-      );
-      await patchVersion(payload).unwrap();
-    }
   };
 
   const onTaskDragEnd = async (result: DropWithDestination) => {
     const { source, destination } = result;
-
-    // Backup list that is not mutated, used for reverting action if api request fails
-    const backupLists = copyVersionLists(versionLists);
```
juhoeerola, line 194: One case where the view can get out of sync is when the browser is offline.
The tasks can be moved and the edits stay locally, but obviously don't go through.

Either way the page needs to be refreshed to get up to date data after getting back to online, so I don't think the backup list is strictly necessary.

But in that case we should at least show errors. And in case of error, the `versionLists` could be reset from the last fetched data (e.i. we could run the `useEffect`).

juhoeerola (changes requested):
Looks good 👍, removing the task from previous version in the backend is great solution for this.

I think removing the backup list would require either some way to show invalid state/errors, or using the last fetched data as backup.

juhoeerola (approved):

PR #291, opened 2022-02-23
Add socketIo realtime updates

AleksiKytola:


juhoeerola (changes requested):
Bringing in my (still open) comments from the previous pr.

Reducing the console logging: if not now when should we do it?

server/src/api/versions/versions.controller.ts
```diff
@@ -124,6 +146,14 @@ export const patchVersions: RouteHandlerFnc = async (ctx) => {
       { name: name, sortingRank: sortingRank },
     );
 
+    await emitRoadmapEvent(ctx.io, {
```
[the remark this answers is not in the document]
AleksiKytola: I totally forgot about the logging in frontend, fixing

server/src/api/versions/versions.controller.ts
```diff
@@ -124,6 +146,16 @@ export const patchVersions: RouteHandlerFnc = async (ctx) => {
       { name: name, sortingRank: sortingRank },
     );
 
+    if (patched) {
+      await emitRoadmapEvent(ctx.io, {
```
juhoeerola, line 150: This still won't emit the event if tasks are only reordered. I think the correct place would be in the else-branch below.

Change for maybe another time:

The response has now been either `Version` or `number[]`, which was a bit weird and incorrectly typed in the frontend.
Now this may also touch other versions, so maybe this should just return empty OK response, and the frontend would fetch the new state.

juhoeerola (changes requested):
Apart that one change, this looks good

server/src/api/versions/versions.controller.ts
AleksiKytola: I believe that now when we use rtk-query the patch method invalidates the tag and causes a re-fetch anyway. I'll just have the controller emit the event regardless of if any changes were made. If necessary it can be updated in a future PR

juhoeerola: This would be the correct place if we want to avoid emitting the events on failed patches, I think this would be good to include in this PR as all the other routes do this.
```ts
  if (!updated) {
    return void (ctx.status = 404);
  } else {
    // emit the event here
    return void (ctx.body = updated);
  }
```

 Fixing the payload, and types on frontend should be left for future PR. since that is not related to the changes in this one.

PR #293, opened 2022-02-23
Add Customer step to roadmap creation

IidaKainu:


frontend/src/components/modals/AddRoadmapModal.tsx
```diff
@@ -17,18 +23,87 @@ import {
...
+) =>
+  customers.map((customer) => ({
+    ...customer,
+    representatives: customer.representatives.filter(
+      ({ email }) => email !== member.email,
+    ),
+  }));
+
+const getKey = (person: InviteRoadmapUser | RoadmapCreationCustomer) =>
+  isInviteUser(person) ? person.email : `${person.name}-${person.email}`;
```
juhoeerola, line 48:
```suggestion
  isInviteUser(person) ? person.email : `${person.name}@${person.email}`;
```

`-` is ambiguous separator, as it can appear any number of times on either side, e.g.
name = "a-b", email = "c@d"
name = "a", email = "b-c@d"

`@` would work, as it is guaranteed to appear exactly once in a valid email, also any character that is not allowed in email address would work

frontend/src/components/modals/modalparts/CustomerModalParts.tsx
```diff
@@ -112,3 +122,104 @@ export const SelectRepresentatives: FC<{
...
+        onClose={() => onCloseError()}
+      >
+        <Trans i18nKey="The combination of name and email should be unique" />
+      </Alert>
+      <div className={classes(css.buttons)}>
+        <button
+          className="button-small-filled"
+          type="button"
+          disabled={!customer.name || !customer.email}
+          onClick={() => onSubmit({ ...customer, representatives })}
```
juhoeerola, line 207: At the moment this allows submitting invalid fields. The first click is blocked and the field shows the error status, but the second click succeeds.

This and the member modal also have some other unexpected/undesirable interactions with the step form, for example <kbd>enter</kbd> submits the form, not the nested modal, advancing to the next step.

We should somehow change the submit action to submit the currently open edit modal instead, then the normal form validation will also catch the invalid fields.

frontend/src/utils/CustomerUtils.ts
```diff
@@ -10,7 +14,7 @@ const difference = (first: number, second: number) =>
   180 - Math.abs(Math.abs(first - second) - 180);
 
 export const randomColor = (
-  customers: Customer[] | undefined,
+  customers: Customer[] | RoadmapCreationCustomer[] | undefined,
```
juhoeerola, line 17: This could just take a list of colours, or `{ color: string }[]` which wouldn't require any changes in the callers

server/src/api/roadmaps/roadmaps.controller.ts
```diff
@@ -37,18 +40,74 @@ export const getRoadmapsUsers: RouteHandlerFnc = async (ctx) => {
...
+      }),
+    );
+    return {
+      ...roadmap,
+      customers: await Customer.query(trx)
+        .withGraphFetched('representatives')
+        .where({ roadmapId }),
+      invitations: await Invitation.query(trx)
+        .withGraphFetched('representativeFor')
+        .where({ roadmapId }),
```
juhoeerola, line 109: I don't think we should include the customers and invitations, even though the `Roadmap` type has optional `Customers[]`, those are no longer included in the roadmaps query.

This could just send the `roadmap` in the response

server/src/api/roadmaps/roadmaps.controller.ts
```diff
@@ -37,18 +40,74 @@ export const getRoadmapsUsers: RouteHandlerFnc = async (ctx) => {
...
+        const representedCustomers = customers.filter(({ name, email }) =>
+          represented.some(
+            (rep: any) => rep.name === name && rep.email === email,
+          ),
+        );
+        return invitation
+          .$relatedQuery('representativeFor', trx)
+          .relate(representedCustomers);
+      }),
+    );
```
juhoeerola, line 101: The invitation emails are not sent, that should be done last here, after all the other steps have succeeded

juhoeerola (changes requested):

IidaKainu:
Thank you for the review: fixes added

> This and the member modal also have some other unexpected/undesirable interactions with the step form, for example enter submits the form, not the nested modal, advancing to the next step.
> 
> We should somehow change the submit action to submit the currently open edit modal instead, then the normal form validation will also catch the invalid fields.

I am still thinking about this though

> At the moment this allows submitting invalid fields. The first click is blocked and the field shows the error status, but the second click succeeds.

I noticed customer email fields are not validated. Validation is now added, does it fix the problem of the second click succeeding?

juhoeerola (changes requested):
I think the "nested forms" will need bit more robust approach. There is still some odd interactions/interference e.g.:
- the "next" button submits the inner form (which would be expected from the implementation, but probably surprising for the user)
- the "previous" button still has some weird behaviour when the inner form is open and has invalid field

I think having only one pair of "submit/next"+ "cancel/previous" buttons active (maybe even visible) at a time might be a good option. I'm not sure how that would be implemented cleanly though.

The server side looked good apart from the two queries, and other aspects of the frontend seem fine

IidaKainu:
Thanks for catching the weird interactions with the innerForm. 

In my opinion disabling the next-button corresponds with the current workflow. Currently the button is disabled if incomplete or incorrect fields exist. Having the inner form open indicates (to me at least) same type of "in progress".
Of course, previous steps can be always navigated to. In that case, the inner form should be closed.

The implementation is not the cleanest though... it's up to the developer to use onPreviousStepClick (or disable other buttons in the step if needed).



Added the change requests 👍

frontend/src/icons/alert_icon_small.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
```
juhoeerola, line 1: Is this actually used anywhere? I could not find any reference to this

juhoeerola (approved):
I agree the implementation could be improved on, but I think it is good enough for now, as nested forms should not be the common use case.

The interactions and logic of the form feels good now, and I didn't find any bugs or surprising behaviour.

If the svg *is* unused, it should be either used or removed, otherwise LGTM 👍

frontend/src/icons/alert_icon_small.svg
IidaKainu: Oh right it was used when customers needed to have representatives. Not needed anymore 👍

PR #296, opened 2022-02-24
Add CDK project for additional staging environment

AleksiKytola:
Diffi on aika ikävä koska copypastettu deployment-projekti, paikat mitä originaali deploymentistä tarvi muuttaa:

`backend-stack.ts`
Secrettien nimet eri
Load balancerilla ei saanu olla saman niminen targetGroup

`
codedeploy-staging-cd.yml`
Deployaa eri urliin, ja buildatessa eri env frontille
Deployaa eri codedeploy applikaation nimellä että se menee stagingiin

`start-server.sh`
Lukee deployment groupin nimen ja looginen branchi kummat secretit laitetaan backendin .enviin

`deployment.ts`
Eri domain yms parametrit
envVarsSecretARN osottaa eri secretiin mikä tehty käsin consolesta
(edited)

`deployment-staging/.gitignore`
poistettu .tsconfig ignoresta, jostain syystä pudotti sen tässä vaikka edellisessä toimi :thinking:

### DIFF
--- a/.github/workflows/codedeploy-staging-cd.yml
+++ b/.github/workflows/codedeploy-staging-cd.yml
@@ -3,8 +3,7 @@ name: AWS Codedeploy CD
 # Controls when the action will run.
 on:
   push:
-    branches: [staging]
-
+    branches: [staging-sandbox]
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
@@ -24,7 +23,7 @@ jobs:
       - name: Build & Generate frontend deployment package
         run: echo $REACT_APP_API_BASE_URL && cd frontend && yarn && yarn build && cd ..
         env:
-          REACT_APP_API_BASE_URL: https://api.test.roadmapper.vincit.fi/
+          REACT_APP_API_BASE_URL: https://api.staging.roadmapper.vincit.fi/
           CI: false
 
       - name: Configure AWS credentials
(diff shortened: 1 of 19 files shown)


.github/workflows/codedeploy-staging-cd.yml
```diff
@@ -0,0 +1,47 @@
+name: AWS Codedeploy CD
+
+# Controls when the action will run.
+on:
+  push:
+    branches: [master]
```
juhoeerola, line 6:
```suggestion
    branches: [master]
    tags: ['test/**']
```

Maybe we could include an alternative for master branch, like tags (or alternative branch) pattern. The suggestion would match any tag starting with `test/`.
This would allow pushing a tag for testing some unfinished or otherwise experimental branches/commits, that are not ready for merging to master.

This would obviously require coordination between the team, so we don't all try to push tags to different places and pushing to master at the same time. Using the tags would be the special case: when using a tag, let others know so they don't push new code to master.

juhoeerola (approved):
I'd say this is good to go, unless you would like to refactor the duplication before merging

.github/workflows/codedeploy-staging-cd.yml
AleksiKytola: Now that I think about it I think I'll rather make the sandbox environment ONLY deploy on pushes to a staging-sandbox branch as it communicates that there is only one branch and one environment and eliminates some need for developer communication. This also stops any merge into master from interfering with someones testing in a `test/**` commit due to the master deployment potentially overwriting the branch being tested

juhoeerola: This is fine also, but it still requires coordination in the same situations. The coordination would just be: who will push to the staging branch.

The main benefit would be, that it's easier to see what is running in the staging.

PR #298, opened 2022-02-24
Separate CI-workflows for frontend and server

juhoeerola:
Most pull requests only change one or the other, so it would make sense
to run only the required tests. When there are changes to the shared
files, both workflows will run.

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -1,0 +1,25 @@
+name: Frontend tests
+on:
+  pull_request:
+    paths-ignore:
+      # exclude files under /server instead of only including /frontend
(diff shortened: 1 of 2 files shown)


AleksiKytola:
Looking good, this should probably set the CI: false environment variable for the yarn build step so the warnings wont stop it from passing

juhoeerola:
Added the CI=false for frontend.
Would this also be needed for server? That has been fine so far without it.

AleksiKytola:
> Added the CI=false for frontend. Would this also be needed for server? That has been fine so far without it.

Lets only add later for the backend if it becomes necessary

AleksiKytola (approved):
Assuming theres no conflicts with the new staging environment actions, this one is good to go :+1:

PR #299, opened 2022-02-28
Mui update

PatricKangasmaki:
Updated Mui from v4 to v5 to fix the issues that were present in the tooltip components of the v4. (Each time you hovered tooltip, deprecated error would show up in console)

This PR focuses only on the tooltips, do you think it would be worth to update all components that use v4 components and icons so we could get rid of the old package or just update places that might cause issues? I haven't found anything else than tooltip so far that gives errors.


AleksiKytola (approved):

AleksiKytola:
I think you could try and update all the other components to mui v5 as well. It might even be that a simple mass search&replace covers most of them if we're lucky.

PR #300, opened 2022-03-01
Add middleware to check unauthenticated requests

juhoeerola:
This adds a middleware that clears the state on any 401 response that
comes through the redux actions.

There are few functions in the api that are called directly. These are
obviously not handled by this middleware.


PatricKangasmaki (approved):
👍

PR #303, opened 2022-03-03
Update all Mui v4 components to v5

PatricKangasmaki:
Removed old v4 packages and replaced them with v5. Updated all the components to use the new packages. 

Things that broke and needed fixing:

- Roadmap selector widget down-arrow moved into a weird position.
- Planner client weights slider needed reworking, old styling didn't work with v5
- InfoModal HelpIcon styles broke, needed to add higher specificity for the styles to apply


juhoeerola (changes requested):
Looks good 👍 

I found just one possible problem:
> You are loading @emotion/react when it is already loaded.

We seem to have two versions in yarn.lock 11.8.1 directly, and 11.1.1 for react-select. If both could cope with the same version, that would be great.

PatricKangasmaki:
> We seem to have two versions in yarn.lock 11.8.1 directly, and 11.1.1 for react-select. If both could cope with the same version, that would be great.

I downgraded the emotion installed for the Mui from 11.8.1. to 11.1.1. and everything works smoothly

juhoeerola (approved):

PR #307, opened 2022-03-04
Submit innerForm on enter

IidaKainu:


PatricKangasmaki (approved):
Working as intended 👍

PR #308, opened 2022-03-07
Correctly display unrated tasks count in team members list

AleksiKytola:
Previously the row displayed the count of tasks filtered by `isUnrated`, effectively displaying the # of unrated tasks missing from representatives (the count that is used in PO's dashboard and tasklist) instead of the count of the tasks unrated specifically by the PO themself.

### DIFF
--- a/frontend/src/utils/SortRoadmapUserUtils.ts
+++ b/frontend/src/utils/SortRoadmapUserUtils.ts
@@ -1,6 +1,6 @@
 import { RoadmapUser, Customer, Task } from '../redux/roadmaps/types';
 import { RoleType } from '../../../shared/types/customTypes';
-import { unratedTasksAmount } from './TaskUtils';
+import { unratedTasksByUserCount } from './TaskUtils';
 import { SortBy, sortKeyLocale, sortKeyNumeric } from './SortUtils';
 
 export enum UserSortingTypes {
@@ -12,7 +12,6 @@ export enum UserSortingTypes {
 export const userSort = (
   roadmapId?: number,
   tasks?: Task[],
-  users?: RoadmapUser[],
   customers?: Customer[],
 ) => (type: UserSortingTypes | undefined): SortBy<RoadmapUser> => {
   switch (type) {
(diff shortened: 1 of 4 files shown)


juhoeerola (changes requested):
The old count is also used in the sorting. This should be ok to just update, as there are no other users for it

https://github.com/Vincit/VISDOM-Roadmapper/blob/4a7a85c95788d715cf7c27a55765cce6b614a47f/frontend/src/utils/SortRoadmapUserUtils.ts#L23-L28

AleksiKytola:
> The old count is also used in the sorting. This should be ok to just update, as there are no other users for it
> 
> https://github.com/Vincit/VISDOM-Roadmapper/blob/4a7a85c95788d715cf7c27a55765cce6b614a47f/frontend/src/utils/SortRoadmapUserUtils.ts#L23-L28

Separated the unrated tasks count into its own util and used it in both places now

frontend/src/utils/SortRoadmapUserUtils.ts
```diff
@@ -21,10 +21,10 @@ export const userSort = (
     case UserSortingTypes.SORT_ROLE:
       return sortKeyLocale((user) => RoleType[user.type]);
     case UserSortingTypes.SORT_UNRATED:
-      return roadmapId === undefined
+      return !roadmapId || !tasks || !customers
         ? undefined
         : sortKeyNumeric((user) =>
-            unratedTasksAmount(user, roadmapId, tasks ?? [], users, customers),
+            unratedTasksByUserCount(tasks, user, roadmapId, customers),
```
juhoeerola, line 27:
```ts
export const userSort = (
  roadmapId?: number,
  tasks?: Task[],
  customers?: Customer[],
) => (type: UserSortingTypes | undefined): SortBy<RoadmapUser> => {
```

Now the `users` parameter is unused, and should be removed

juhoeerola (changes requested):

juhoeerola (approved):

PR #309, opened 2022-03-07
Upgrade server dependencies

juhoeerola:
Major version upgrade of Objection and Knex.
- the types are now more accurate for queries: query returning a single row (e.g. `.first()` or `.findById()`) now includes `undefined` as possible return type
- json schema format validation is no longer automatically included in `Objection.Model`, the `BaseModel` adds explicitly specified validators back.

Minor version upgrades on other dependencies to fix audit issues.


AleksiKytola (approved):

juhoeerola:
I'll squash these to 3 commits: the last two individually, but all the rest are related to knex & objection upgrade and the intermediate commits are not in working state

PR #313, opened 2022-03-08
Add virtualization to SortableTaskList

IidaKainu:


frontend/src/components/SortableTaskList.tsx
```diff
@@ -28,6 +36,45 @@ export const SortableTaskList: FC<{
...
+      <SortableTask
+        style={{ ...rest }}
+        key={data[index].id}
+        task={data[index]}
+        index={index}
+        disableDragging={disableDragging}
+        showRatings={!!showRatings}
+      />
+    );
+  };
```
IidaKainu, line 76: This is to avoid weird error msg: https://github.com/atlassian/react-beautiful-dnd/issues/1647#issuecomment-686956052

frontend/src/components/SortableTaskList.module.scss
```diff
@@ -34,3 +34,20 @@
...
+  word-wrap: break-word;
+
+  -webkit-box-orient: vertical;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+}
+
+.measureTaskName {
+  @extend .taskName;
+  visibility: hidden;
```
juhoeerola, line 52: This seems to have less width available (about 163px)

This leads to the gaps closing sometimes when the name length falls within that difference

frontend/src/components/SortableTaskList.module.scss
```diff
@@ -34,3 +34,20 @@
   position: sticky;
   top: 0;
 }
+
+.taskName {
```
juhoeerola, line 38: This and `.measureTaskName` are same as in TaskMap.module.scss, apart from the additional dragging styles.
Maybe these could be defined at one place

juhoeerola (changes requested):
I didn't notice much difference in the loading time (when opening the list) compared to the previous, but it's a bit nicer now that all tasks are restricted to couple lines max.

I noticed the width wasn't quite right for the calculations.

And also, a previous style bug: the `.info` css class should be defined in TaskMap.module.scss, not TaskMapPage.module.scss. I think this slipped through in the refactor

IidaKainu:
Good catches: fixed the name measuring, moved the tasknapshared styles to shared.scss and .info to TaskMap.module.scss

frontend/src/components/TaskMap.module.scss
```diff
@@ -7,16 +7,15 @@ $ZOOM: var(--zoom, 1);
...
+  width: 16px;
+}
 
-  -webkit-box-orient: vertical;
-  display: -webkit-box;
-  -webkit-line-clamp: 2;
+.taskName {
+  @extend .taskName;
 
   &.dragging {
```
juhoeerola, line 20: I'm not too familiar with the name resolution/scoping and how exactly it would work here,
but even if this works, I'd pick different names just for clarity

Or would it be sufficient to add just the dragging rule? (the dragging styles aren't exactly needed for the measuring)

```suggestion
.taskName.dragging {
```

juhoeerola (commented):
Looking good 👍 

I just got a bit weirded out by the one css rule...

frontend/src/components/TaskMap.module.scss
IidaKainu: Valid point even though it works in development. Let's go with a different name for clarity

IidaKainu:
Different css-name added 👍

juhoeerola (approved):

PR #315, opened 2022-03-09
Fix links in emails

juhoeerola:
Two emails had different link in the button and as text.
I extracted the link to a variable to avoid the mistake in the future.

Seems like I also missed one broken icon link when reviewing these.


PatricKangasmaki (approved):
Great catches 👍

PR #316, opened 2022-03-09
Customer ratings without user or with old representatives

juhoeerola:
Fixes the handling of ratings user has given for a customer the user no longer represents.

Correctly shows ratings from users that have left the roadmap.

Ratings for a customer are already deleted, as they should, when the associated customer is deleted.

Todo for another issue: decide what should happen to ratings when the associated user account is deleted.


AleksiKytola (approved):
I didnt do very extensive testing through the UI itself. Code seems good to go :+1:

PR #317, opened 2022-03-09
Bootstrap to mui

IidaKainu:


IidaKainu:
Good catches: removed the info icon and centered the selector widget icon. I could not find any other broken styles

juhoeerola (approved):
Didn't see anything else broken either 👍

PR #318, opened 2022-03-09
Add taskmap handle style effects

IidaKainu:
https://trello.com/c/nTXYc0AF/365-task-map-handleihin-onclick-hoverille-effektit
It looks like this may be a expired task. If this is not an improvement, lets close the pr

- Add hover effect to handles (the same effect is preserved when user starts connecting from the handle)
- Unconnectable handles are grayed out
- Connectable handles have a hover effect


juhoeerola (approved):
👍

PR #319, opened 2022-03-09
Fix StepForm bug

IidaKainu:
Fix bug: if user presses "No, go back" in the cancel confirmation, the
application would crash due to a missing undefined check.


juhoeerola (approved):
👍

PR #320, opened 2022-03-10
Make cursor not-allowed when hovering groups where task cannot be dro…

PatricKangasmaki:
Cursor is now type "not-allowed" when hovering task groups that aren't allowed

### DIFF
--- a/frontend/src/components/TaskMapEdge.tsx
+++ b/frontend/src/components/TaskMapEdge.tsx
@@ -1,14 +1,28 @@
 import { FC, MouseEvent, useState, useEffect, useCallback } from 'react';
 import { useSelector } from 'react-redux';
 import classNames from 'classnames';
-import { getMarkerEnd, getBezierPath } from 'react-flow-renderer';
+import {
+  getMarkerEnd,
+  getBezierPath,
+  Position,
+  ArrowHeadType,
+} from 'react-flow-renderer';
 import { chosenRoadmapIdSelector } from '../redux/roadmaps/selectors';
 import css from './TaskMapEdge.module.scss';
 import { apiV2 } from '../api/api';
 
 const classes = classNames.bind(css);
 
-const DrawPath: FC<any> = ({
+interface DrawPathProps {
+  id: string;
+  d: string;
+  markerEnd: string;
+  disableInteraction: any;
+  isLoading: any;
+  setIsLoading: any;
+}
+
+const DrawPath: FC<DrawPathProps> = ({
   id,
   d,
   markerEnd,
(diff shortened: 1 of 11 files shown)


frontend/src/components/TaskMapTask.module.scss
```diff
@@ -48,6 +48,10 @@
   &.loading:not(.dragging, .draggingOutside) {
     cursor: wait;
   }
```
juhoeerola, line 50: Looking at this, the loading cursor style should also be applied to all the cases where the `not-allowed` is (the group, and the handles), as it now doesn't seem to show consistently.

juhoeerola (approved):
Looking good 👍 

The loading cursor might be good to fix now at the same time, as it is needed in all the same places. Easier now than hunting those down later.

frontend/src/components/TaskMapEdge.tsx
```diff
@@ -8,7 +8,14 @@ import { apiV2 } from '../api/api';
 
 const classes = classNames.bind(css);
 
-const DrawPath: FC<any> = ({ id, d, markerEnd, disableInteraction }) => {
+const DrawPath: FC<any> = ({
```
juhoeerola, line 11: Some types might be good to add for this (as well as `CustomEdge`'s props)

frontend/src/components/TaskMapTask.tsx
```diff
@@ -26,7 +26,10 @@ export interface TaskProps {
   setSelectedTask?: any;
   checked: { from: boolean; to: boolean };
   disableDragging: boolean;
+  dropDisabled?: boolean;
+  setGroupDraggable: any;
```
juhoeerola, line 30:
```suggestion
  setGroupDraggable: (_: boolean) => unknown;
```

Could be a suitable type, since that is what it is used for

frontend/src/components/TaskMapTask.tsx
```diff
@@ -140,8 +157,18 @@ const SingleTask: FC<
...
-> = ({ taskId, index, disableDragging, ...rest }) => (
+> = ({
+  taskId,
+  index,
+  disableDragging,
+  dropDisabled,
+  setGroupDraggable,
+  draggingSomething,
+  isLoading,
+  ...rest
```
juhoeerola, line 170:
```suggestion
  ...rest
```

These could be included in the rest, as these are only forwarded

frontend/src/components/TaskMapTask.tsx
```diff
@@ -155,6 +182,10 @@ export const DraggableSingleTask: FC<
           provided={provided}
           snapshot={snapshot}
           disableDragging={disableDragging}
+          dropDisabled={dropDisabled}
+          setGroupDraggable={setGroupDraggable}
+          draggingSomething={draggingSomething}
+          isLoading={isLoading}
           {...rest}
```
juhoeerola, line 189:
```suggestion
          {...rest}
```

frontend/src/components/TaskMapTask.tsx
```diff
@@ -140,8 +157,18 @@ const SingleTask: FC<
 export const DraggableSingleTask: FC<
   TaskProps & {
     index: number;
+    draggingSomething: boolean;
```
juhoeerola, line 160: This might fit better in `TaskProps`, since it is always used with that type

PatricKangasmaki:
Great points, added all your suggestions 👍 I also fixed the bug where not-allowed cursor didn't show up correctly when hovering over a single task (the bug which showed up in the demo). Css was just missing one class from the :not pseudo class

`&.draggingSomething:not(.loading) { ...` to
`&.draggingSomething:not(.loading, .dropDisabled) { ...`

juhoeerola (approved):
👍 seems to work fine (also on safari)

PR #323, opened 2022-03-11
Fix user update after editing a customer

juhoeerola:
The new user info was fetched too early immediately after starting the
patch query. Now the fresh user info is fetched only after the patch has
succeeded.


AleksiKytola (approved):
:+1: . This led me to think whether we properly update userInfo via the websocket if the PO edits their roles for example.

juhoeerola:
> whether we properly update userInfo via the websocket if the PO edits their roles for example.

I don't think we do. I'm also not sure if it would be enough to emit the event from the appropriate server routes, or does that get out of sync for the userInfo state.

PR #325, opened 2022-03-14
Fixed some text in landingpage

DanielMakinen:
Small fix. Time estimation was written twice in the feature-list.

### DIFF
--- a/frontend/src/pages/LandingPage.tsx
+++ b/frontend/src/pages/LandingPage.tsx
@@ -191,11 +191,6 @@ export const LandingPage = () => {
                 tasks by how complex they see each of them, and the customers by
                 how much value it brings to them.
               </li>
-              <li>
-                <strong>Time estimation</strong> - Based on pure guesses or
-                already realized milestones, PO can estimate how other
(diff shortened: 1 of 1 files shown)


PatricKangasmaki (approved):
👍

juhoeerola:
Just a note on the commit message style: It's best to keep the subject line short and to the point. There is plenty of space to elaborate in the body paragraphs. Also present, instead of past, tense is usually preferred for the subject line ("Fix" instead of "Fixed").

For example:

```
Remove duplicate paragraph from landing page

More explanation here if needed, multiple paragraphs is fine. Wrap
lines at appropriate length, so they stay nice to read.
```

PatricKangasmaki:
Good point from @juhoeerola, here's a good read about git commits as well (though juho already explained the key points)
https://cbea.ms/git-commit/

juhoeerola:
There is also a great git course in univincity that goes to many other aspects of using git as well

DanielMakinen:
Thanks for feedback! I read the article and will later go through the univincity git-course.

I changed the commit message to be more concise.

PR #326, opened 2022-03-15
Fix incorrect permission checks for editing tasks

DanielMakinen:
TaskOverview page used to check for wrong permissions for task editing.

Also added typing for "err" in try-catch cause eslint was complaining about it.


juhoeerola (approved):
👍

PR #327, opened 2022-03-15
Add drag-n-drop instructions to MilestonesEditor

IidaKainu:
If no milestones have tasks, display a drag-n-drop info message in all milestones.

The branch name is a bit outdated, it was decided that all milestones should have the instruction in this case.

### DIFF
--- a/frontend/src/i18/english.ts
+++ b/frontend/src/i18/english.ts
@@ -72,6 +72,7 @@ export const english = {
     'Rate customer': 'Rate customer',
     milestones: 'milestones',
     Milestones: 'Milestones',
+    'Milestone task instructions': 'Drag & drop tasks here',
     'Add milestone': 'Add milestone',
     'Milestone name': 'Milestone name',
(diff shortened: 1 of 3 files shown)


juhoeerola (approved):
I think it would be fine, even helpful \*, to show the instruction text placeholder on any empty milestone.
And I don't think it's too intrusive or annoying. But those are somewhat personal preference.

Either way looks good 👍 

\* lets say you come into a project that already has milestones, and you should create a new one, but you haven't had the chance to edit milestones prior to that so you have no idea how the tasks ended up in those existing milestones

IidaKainu:
> * lets say you come into a project that already has milestones, and you should create a new one, but you haven't had the chance to edit milestones prior to that so you have no idea how the tasks ended up in those existing milestones

Valid case where the user could benefit from instructions. I agree it would not be too annoying as a project would rarely have empty milestones. Changed show in empty milestones, merging 👍

PR #328, opened 2022-03-15
Add color-forest hover effect to expand icons

AnttiVaarala:
Change color of expand more and expand less icons to $COLOR-FOREST while hovering over them

### DIFF
--- a/frontend/src/components/ExpandableColumn.module.scss
+++ b/frontend/src/components/ExpandableColumn.module.scss
@@ -27,6 +27,9 @@
   svg {
     margin-right: 11px;
   }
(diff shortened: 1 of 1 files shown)


PatricKangasmaki:
Looking good 👍 Same comment here as Juho gave you in the other PR about commit practices https://github.com/Vincit/VISDOM-Roadmapper/pull/325#issuecomment-1067668714. Currently the message seems a bit long

Shortened example:

```
Add color-forest hover effect to expand icons

Change color of expand more and expand less icons to $COLOR-FOREST while hovering over them
```

PR #330, opened 2022-03-16
Cleanup socketIo code, add event for refetching userInfo on role updates

AleksiKytola:


AleksiKytola:
Userinfo updates seemed to work in my testing. Swapping an admin user into a developer seamlessly transitioned their UI into "developer view" etc.

server/src/api/customer/customer.controller.ts
```diff
@@ -87,13 +96,32 @@ export const patchCustomer: RouteHandlerFnc = async (ctx) => {
...
         .unrelate();
+
+      for (let representative of removed) {
+        await emitRoadmapEvent(ctx.io, {
+          roadmapId: Number(ctx.params.roadmapId),
+          onlyEmitToUserId: Number(representative),
+          event: ClientEvents.USERINFO_UPDATED,
+          eventParams: [],
+        });
+      }
```
juhoeerola, line 107:
```suggestion
      await emitRoadmapEvent(ctx.io, {
        roadmapId: Number(ctx.params.roadmapId),
        onlyEmitToUserIds: removed,
        event: ClientEvents.USERINFO_UPDATED,
        eventParams: [],
      });
```

Having the "only id" be an array of ids instead would make this a lot nicer.
Also each iteration of the loop needs to do pretty much the same things,
so there will be a lot of repeated work that.

This also reduces the excessive `await`ing

Sending to a single user still very reasonable, just needs wrapping the one id to an array.

server/src/utils/socketIoUtils.ts
```diff
@@ -61,20 +64,23 @@ export const setSubscribedRoadmap = (socket: ExtendedSocket) => async (
...
 };
 
 export type EmitRoadmapEventParams<T extends ClientEvents> = {
   roadmapId: number;
-  dontEmitToUserId?: number;
-  requirePermission?: number;
   event: T;
   eventParams: Parameters<ClientEventsMap[T]>;
+  dontEmitToUserId?: number;
+  onlyEmitToUserId?: number;
```
juhoeerola, line 82:
```suggestion
  onlyEmitToUserIds?: number[];
```

This could be an array

server/src/utils/socketIoUtils.ts
```diff
@@ -83,27 +89,24 @@ export const emitRoadmapEvent = async <T extends ClientEvents>(
...
+    onlyEmitToUserId,
     eventParams: parameters,
   } = { ...emitRoadmapEventParams };
 
-  console.log('Emitting roadmap socket event:');
-  console.log(JSON.stringify(emitRoadmapEventParams, undefined, 2));
-
   const roomName = `roadmap-${roadmapId}`;
   const roadmapSockets = await io.in(roomName).fetchSockets<ISocketData>();
   const socketUserIds = roadmapSockets.map((s) => s.data.user.id);
```
juhoeerola, line 101:
```suggestion
  let socketUserIds = roadmapSockets.map((s) => s.data.user.id);
  if (dontEmitToUserId !== undefined) {
    socketUserIds = socketUserIds.filter((id) => id !== dontEmitToUserId);
  }
  if (onlyEmitToUserIds) {
    socketUsers = socketUsers.filter((user) => onlyEmitToUserIds.includes(user.id));
  }
```

It would make sense to filter out only the requested ids as soon as possible, and that would be here.

server/src/utils/socketIoUtils.ts
```diff
@@ -83,27 +89,24 @@ export const emitRoadmapEvent = async <T extends ClientEvents>(
...
-  const socketUsers = await User.query()
+  let socketUsers = await User.query()
```
juhoeerola, line 102:
```suggestion
  const socketUsers = await User.query()
```
then this could be const

server/src/utils/socketIoUtils.ts
```diff
@@ -116,6 +119,5 @@ export const emitRoadmapEvent = async <T extends ClientEvents>(
 
     // Sockets are added to a room with their associated user id when they connect to allow sending like this
     io.to(`${user.id}`).emit(event, ...parameters);
```
juhoeerola, line 121:
```suggestion
    io.to(usersWithPermission.map((user) =>`${user.id}`).emit(event, ...parameters);
```

the `.to()` method can also take an array, I'm not sure if it would have benefits, but it would be an option.

juhoeerola (changes requested):
Updates seem to work well

Few suggestions to improve the emitting

AleksiKytola:
I converted both dontEmitToUserIds and onlyEmitToUserIds to arrays now.

server/src/utils/socketIoUtils.ts
juhoeerola: Is there some reason why the ids could or should not be filtered here, but only after fetching from the database?

AleksiKytola: No reason at all, a brainfart on my part. I thought I had added it before the query :sweat_smile:

PR #331, opened 2022-03-16
Unify link stylings

DanielMakinen:
Links retain their color when being hovered over.

### DIFF
--- a/frontend/src/pages/ForgotPasswordPage.tsx
+++ b/frontend/src/pages/ForgotPasswordPage.tsx
@@ -163,7 +163,10 @@ export const ForgotPasswordPage = () => {
             <Trans i18nKey="We have sent an email" values={{ email }} />
           </p>
         </div>
-        <div className="formFooter" style={{ textAlign: 'center' }}>
+        <div
+          className={classes(css.formFooter)}
+          style={{ textAlign: 'center' }}
+        >
           <div style={{ marginBottom: '16px' }}>
             <Trans i18nKey="Wrong email address?" />{' '}
             <button
--- a/frontend/src/shared.scss
+++ b/frontend/src/shared.scss
@@ -418,7 +418,6 @@ textarea {
   background-color: transparent;
   padding: 0;
   border: none;
-  font-weight: bold;
 
   &:disabled {
     color: $COLOR_BLACK40;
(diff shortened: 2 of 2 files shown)


PatricKangasmaki:
> Links retain their color when being hovered over.  

This seems to work well 👍 This could also be added here:

- toinen on boldattu, toinen ei

Currently link-buttons are bold by default, regular links aren't. Note that you'll need to apply semi-bold (font-weight: 600) to links in Reset link sent page (path: ForgotPasswordPage.tsx -> linkSentView -function) manually. I didn't find anything else where changing the default font-weight would cause issues but you can check as well if you feel like its needed.

DanielMakinen:
Don't mind the force-push spam, I was having a stroke.

Thanks for pointing out that linkButtons need use font-weight:600. I need to learn to check Figma for these things.

I removed margin-left from footer links, as it was incorrectly defined without "px" suffix and we were already using single space to separate links from text.

So now linkButtons don't have boldness by default. I defined their boldness in modals with .formFooter class like with normal links.

I'm really curious why there are linkButtons in the first place. Are these just some old artifacts or can they do something React-Router Links can't?

PatricKangasmaki (approved):
Looking good 👍 

> I'm really curious why there are linkButtons in the first place. Are these just some old artifacts or can they do something React-Router Links can't?

Links from React-Router are basically just hyperlinks (`<a href='' ...>`) so they're used in places that act just like regular links that redirect you somewhere. You'll find 99% of the links defined like `<Link to="/somepath" />` which would translate to `<a href="/somepath">`.  Sometimes they have small onClicks as well, such as closing the modal.

LinkButtons on the other hand, they're regular buttons with onClicks such as `<button className={classes(css.linkButton, css.green)} onClick={() => handleSelect(user.defaultRoadmapId === roadmapId)} ...>` and the `handleSelect` is a complicated multiline function. They just perform some kind of an action and they will not redirect you anywhere. I'm not sure about the origin of the linkButtons and how they were originally designed to work, but this is how they are currently. There aren't many places where they used anymore though.

In short, React-Router Links are hyperlinks and they're used in redirecting, linkButtons are styled buttons that have some onClick effects but they don't redirect you anywhere but they both look the same.

DanielMakinen:
Thanks for the explanation!

Is there something preventing putting bigger multiline functions on RouterLinks onClick? If not, then couldn't you in theory use link as a button if you just use <Link to="#" onClick={() => someHugeFunction()}>Click me</Link>? Assuming you want the styling to be like with links, as with linkButtons.

DanielMakinen:
Makes sense, thanks for the thorough explanation!

I need to do some reading about ARIA.

PR #335, opened 2022-03-17
Percentage bar component for client value

juhoeerola:


AleksiKytola:
I noticed on my screen (thats small enough horizontally that I get a scrollbar) when scrolling the background lines with the axis notation dont scroll together with the bars causing them to disconnect
image

AleksiKytola:
A small CSS detail. User-select: none or such should be added, currently chrome renders the flashing typing cursor on the colored divs of the bars when I click on them (generally chrome shows this cursor in a ton of weird places though)
image

AleksiKytola:
Also not a very big thing but I think It'd be neat if the customer color dots to appeared in front of the labels also, ie.
<Dot/> CUSTOMER 1: 40.6%
<Dot/> CUSTOMER 2: 43.8%
<Dot/> CUSTOMER 3: 15.6%

AleksiKytola:
image
this could also be added in this PR while we're at it

juhoeerola:
- Positioning fixed
- added `user-select: none`, but could not verify, as I couldn't get that showing in the first place
- changed the title to one from the figma proto: "Client shares of the total value"
- implemented the new tooltip design
  - I changed the title to just "Client shares" as those are not weighted.
  - I didn't show the percentage bar another time in the tooltip, as it would be exactly the same as the user is hovering over 
  - didn't implement the "too far off" calculation, but added a field to `CustomerStakes` for it

AleksiKytola:
Works perfectly for me now and looks great! :+1:

PR #337, opened 2022-03-18
Allow business representatives to create tasks

AleksiKytola:


shared/types/customTypes.ts
```diff
@@ -60,6 +60,9 @@ export enum RoleType {
     Permission.VersionRead |
     Permission.RoadmapReadUsers,
   Business = Permission.TaskRead |
+    Permission.TaskCreate |
```
juhoeerola, line 63: This would need a corresponding migration to update the existing business users in the database.

juhoeerola (changes requested):
Noticed two bugs, otherwise looks good

shared/types/customTypes.ts
AleksiKytola: A good catch

frontend/src/components/TopBar.tsx
```diff
@@ -34,7 +34,7 @@ export const TopBar: FC<{
       />
       {children}
       <div className={classes(css.rightSide)}>
-        {role === RoleType.Admin && (
+        {(role === RoleType.Admin || role === RoleType.Business) && (
```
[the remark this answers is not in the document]
AleksiKytola: I non-optionally parametrized the roles that the buttons will be shown to.

shared/types/customTypes.ts
AleksiKytola: Migration added. I suppose I Cant directly use RoleType.Business in the migration as future changes to that type would screw up the old migrations, so I just have 
```

const oldBusinessRole =
  Permission.TaskRead |
  Permission.TaskEdit |
  Permission.TaskDelete |
  Permission.TaskRate |
  Permission.TaskRatingEdit |
  Permission.TaskValueRate |
  Permission.CustomerRepresent |
  Permission.RoadmapReadUsers;

const newBusinessRole = oldBusinessRole | Permission.TaskCreate;

```

PR #338, opened 2022-03-18
Close ExpandableColumn on render if empty

DanielMakinen:
ExpandableColumn renders untoggled if there are no unordered tasks.

### DIFF
--- a/frontend/src/pages/MilestonesEditor.tsx
+++ b/frontend/src/pages/MilestonesEditor.tsx
@@ -67,6 +67,7 @@ export const MilestonesEditor = () => {
   const dispatch = useDispatch<StoreDispatchType>();
   const [versionLists, setVersionLists] = useState<VersionListsObject>({});
   const [expandUnordered, setExpandUnordered] = useState(true);
+  const [areInitialVersions, setAreInitialVersions] = useState(true);
 
   useEffect(() => {
     if (!roadmapsVersions) return;
@@ -86,8 +87,13 @@ export const MilestonesEditor = () => {
       SortingOrders.DESCENDING,
     )(Array.from(ratedTasks.values()));
 
+    if (areInitialVersions) {
+      setAreInitialVersions(false);
+      if (!newVersionLists[ROADMAP_LIST_ID].length) setExpandUnordered(false);
+    }
+
     setVersionLists(newVersionLists);
-  }, [customers, tasks, roadmapsVersions, isError]);
+  }, [customers, tasks, roadmapsVersions, isError, areInitialVersions]);
 
   const addVersion = () => {
     dispatch(
--- a/frontend/src/pages/MilestonesEditor.tsx
+++ b/frontend/src/pages/MilestonesEditor.tsx
@@ -29,6 +29,7 @@ import { move } from '../utils/array';
 import { InfoTooltip } from '../components/InfoTooltip';
 import css from './MilestonesEditor.module.scss';
 import { apiV2 } from '../api/api';
+import { LoadingSpinner } from '../components/LoadingSpinner';
 
 const classes = classNames.bind(css);
 
@@ -329,7 +330,7 @@ export const MilestonesEditor = () => {
     </Droppable>
   );
 
-  return (
+  const renderPage = () => (
     <DragDropContext onDragEnd={onDragEnd}>
       <InfoTooltip
         title={
@@ -370,4 +371,12 @@ export const MilestonesEditor = () => {
       </div>
     </DragDropContext>
   );
+
+  const renderLoading = () => (
+    <div className={classes(css.loadingSpinnerContainer)}>
+      <LoadingSpinner />
+    </div>
+  );
+
+  return areInitialVersions ? renderLoading() : renderPage();
 };
--- a/frontend/src/pages/MilestonesEditor.tsx
+++ b/frontend/src/pages/MilestonesEditor.tsx
@@ -70,6 +70,12 @@ export const MilestonesEditor = () => {
   const [areInitialVersions, setAreInitialVersions] = useState(true);
 
   useEffect(() => {
+    if (!areInitialVersions || !versionLists[ROADMAP_LIST_ID]) return;
+    setAreInitialVersions(false);
+    if (!versionLists[ROADMAP_LIST_ID].length) setExpandUnordered(false);
+  }, [areInitialVersions, versionLists]);
+
+  useEffect(() => {
     if (!roadmapsVersions) return;
 
     const newVersionLists: VersionListsObject = {};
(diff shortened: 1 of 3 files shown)


frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -86,8 +87,13 @@ export const MilestonesEditor = () => {
       SortingOrders.DESCENDING,
     )(Array.from(ratedTasks.values()));
 
+    if (areInitialVersions) {
+      setAreInitialVersions(false);
+      if (!newVersionLists[ROADMAP_LIST_ID].length) setExpandUnordered(false);
+    }
+
```
IidaKainu, line 94: It's better to separate this into its own useEffect, otherwise this useEffect would run once unnecessarily (when areInitialVersions is set to false)

It would look something like this

`useEffect(() => {
    if (!areInitialVersions || !versionLists[ROADMAP_LIST_ID]) return;
    setAreInitialVersions(false);
    if (!versionLists[ROADMAP_LIST_ID].length) setExpandUnordered(false);
  }, [areInitialVersions, versionLists]);`

IidaKainu (changes requested):
Functions well 👍 Found a problem with the code

IidaKainu:
I added a transition-effect to the ExpandableColumn. It will make the closing of the column look less choppy once the useEffect has been updated 👍

DanielMakinen:
> It's better to separate this into its own useEffect, otherwise this useEffect would run once unnecessarily (when areInitialVersions is set to false)
Thanks for pointing this out, should be fixed now!

Transition works well and looks less choppy. 👍

It bothers me a bit that there is any kind of movement when that page is opened, but i guess there would need to be a really big  change to remedy that. It would need to wait for the API response before rendering anything at all.

Thanks for the help, i guess this can be merged now.

IidaKainu (commented):
> It bothers me a bit that there is any kind of movement when that page is opened, but i guess there would need to be a really big change to remedy that. It would need to wait for the API response before rendering anything at all.

I agree it's a little annoying. Now that I think about it, the loading spinner can be returned while the areInitialVersions is true to prevent the movement. The page is currently white while it's waiting for the response, so the difference would not be drastic. Up to you though 👍

DanielMakinen:
Maybe it is not as big of a task as I initially thought. I look into that and see what i can do.

frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -363,4 +371,12 @@ export const MilestonesEditor = () => {
...
     </DragDropContext>
   );
+
+  const renderLoading = () => (
+    <div className={classes(css.loadingSpinnerContainer)}>
+      <LoadingSpinner />
+    </div>
+  );
+
+  return areInitialVersions ? renderLoading() : renderPage();
```
PatricKangasmaki, line 381: Well you could get rid of the functions `renderPage `and `renderLoading` and simply handle everything in the return. Less code but might be a bit harder to read so might not be worth it though. Maybe if the `areInitialVersions` was called something like `isLoading` here then it could be more readable
```suggestion

  return areInitialVersions ? (
    <div className={classes(css.loadingSpinnerContainer)}>
      <LoadingSpinner />
    </div>
  ) : (
    <DragDropContext onDragEnd={onDragEnd}>
      <InfoTooltip
        title={
          <div>
            <Trans i18nKey="Planner milestones tooltip">
              Milestones create your project’s
              <Link to={`/roadmap/${roadmapId}/planner/graph`}>roadmap.</Link>
            </Trans>
          </div>
        }
      >
        <div className={classes(css.infoBackground)}>
          <InfoIcon className={classes(css.tooltipIcon, css.infoIcon)} />
        </div>
      </InfoTooltip>
      <div className={classes(css.layoutRow, css.overflowYAuto)}>
        <ExpandableColumn
          className={classes(css.unorderedCol)}
          expanded={expandUnordered}
          onToggle={() => setExpandUnordered((prev) => !prev)}
          title={
            <div>
              {`${t('Unordered tasks')} (${
                versionLists[ROADMAP_LIST_ID]?.length ?? 0
              })`}
            </div>
          }
        >
          <SortableTaskList
            listId={ROADMAP_LIST_ID}
            tasks={versionLists[ROADMAP_LIST_ID] || []}
            disableDragging={disableDrag}
            showRatings
            showSearch
          />
        </ExpandableColumn>
        {versionLists && renderMilestones()}
      </div>
    </DragDropContext>
  );
```

IidaKainu: Even better (for readability), return early if areInitialVersions is true
```suggestion
  if (areInitialVersions)
    <div className={classes(css.loadingSpinnerContainer)}>
      <LoadingSpinner />
    </div>;
  return (
    <DragDropContext onDragEnd={onDragEnd}>
      <InfoTooltip
        title={
          <div>
            <Trans i18nKey="Planner milestones tooltip">
              Milestones create your project’s
              <Link to={`/roadmap/${roadmapId}/planner/graph`}>roadmap.</Link>
            </Trans>
          </div>
        }
      >
        <div className={classes(css.infoBackground)}>
          <InfoIcon className={classes(css.tooltipIcon, css.infoIcon)} />
        </div>
      </InfoTooltip>
      <div className={classes(css.layoutRow, css.overflowYAuto)}>
        <ExpandableColumn
          className={classes(css.unorderedCol)}
          expanded={expandUnordered}
          onToggle={() => setExpandUnordered((prev) => !prev)}
          title={
            <div>
              {`${t('Unordered tasks')} (${
                versionLists[ROADMAP_LIST_ID]?.length ?? 0
              })`}
            </div>
          }
        >
          <SortableTaskList
            listId={ROADMAP_LIST_ID}
            tasks={versionLists[ROADMAP_LIST_ID] || []}
            disableDragging={disableDrag}
            showRatings
            showSearch
          />
        </ExpandableColumn>
        {versionLists && renderMilestones()}
      </div>
    </DragDropContext>
  );
```

DanielMakinen: I originally tried to put everything in return with ternary, but prettier-linter insisted that has to be made look super confusing, so I ditched that and made own functions for them.

This is the version that the linter likes.

image

At least to me that looks less clear, but it might look better for others.

And yeah, 'isLoading' could make more sense at this point.

DanielMakinen: Oh, you already posted this version 😄

DanielMakinen: Something like this would be pretty clear to me, but the linter gets super mad. Its not as dense, but I think it shows more clearly what each of these ternary operations results would do.

image

But then again, maybe I'm a freak and this is actually a horrible design 😄

DanielMakinen: Iidas version looks best to me. I don't know why her message took so long to be visible for me.

But thats very readable, probably gonna go with that one.

DanielMakinen: Unless, could it be confusing that the "official" return might not happen at all? Or is that fine?

IidaKainu: > Unless, could it be confusing that the "official" return might not happen at all? Or is that fine?

The same pattern is often used in this project, so it's okay 👍

AleksiKytola: Returning early does have some drawbacks but overall its the pattern I prefer

https://medium.com/swlh/return-early-pattern-3d18a41bba8

IidaKainu (approved):

PR #344, opened 2022-03-22
Fix roadmap updation on invitation acception

IidaKainu:
New roadmap would not appear in the list after accepting an invitation.
Now roadmaps are refetched on acception.


PatricKangasmaki:
Looking good 👍

PatricKangasmaki (approved):

PR #345, opened 2022-03-22
Add disable-effect for unconnectable tasks

IidaKainu:
Now the whole task is disabled (style-wise) instead of only the handle.


PatricKangasmaki (approved):
Looking great 👍👍

PR #349, opened 2022-03-23
Fix taskmap tooltip styling

AnttiVaarala:


frontend/src/components/Overview.tsx
```diff
@@ -78,7 +78,14 @@ export const OverviewContent: FC<OverviewContentProps> = ({
                   format={format}
                 />
               ) : (
-                <div className={classes(css.value)}>{value}</div>
+                <div
+                  className={classes(
+                    css.value,
+                    typeof format !== 'undefined' && css[format],
```
AleksiKytola, line 84: when using the `classNames` library for classnames, you can pass an object for conditional classnames, i.e.
```
    className={classNames({
      [css.value]: true,
      [css.format]: format, // or !!format perhaps to convert to boolean if typescript wants it
    })}
```

AleksiKytola: Also to check for undefined you can just put the value in the place of a boolean expression (like above) 
https://www.sitepoint.com/javascript-truthy-falsy/

unless I missed something and using `  typeof format !== 'undefined' && css[format]` was intentional

PatricKangasmaki: You could even go a bit further with this and change the `[css.value]: true,` to
```
        className={classes(css.value, {
          [css.format]: format,
        })}
```

PatricKangasmaki: This way you can separate the ones that get applied all the time from the conditional ones nicely

PatricKangasmaki:
I would also add some margin on the left and the bottom in the `taskOverviewContainer` on `TaskMapPage.tsx`. To me, something like `  margin-left: 60px; margin-bottom: 15px;` looked good. 60px on the left so it aligns with the unstaged tasks sidebar when it's closed and 15px on the bottom so it would look a bit cleaner since it's not so low anymore.

I didn't add commit about this, I'll let you guys figure it out and see what you like 😄 Also, remember to test how everything looks when the task titles and descriptions are a bit longer and realistic (not these Test task 1 and Test Desc 1).

juhoeerola (changes requested):
I'm slightly sceptical about hard coding the order and number of elements in so may places (`nth-child(1)` and `nth-child(2)`), but I suppose that's ok as long as we remember it is done that way.

juhoeerola (approved):

PR #350, opened 2022-03-23
Add confirmation to change password

DanielMakinen:
So I made the change password modal that is in style with whats on Figma (except for those toggles that let you see passwords in cleartext. Should i make a new task in Trello to add those in this modal and maybe to ResetPassword modal?)

With this change, that EditableTextWithButtons don't really make sense to me. 
image
Should this just be some kind of a button that opens the modal? I temporarily used linkButton for the functionality.

### DIFF
--- a/frontend/src/components/UserInfoCard.module.scss
+++ b/frontend/src/components/UserInfoCard.module.scss
@@ -35,3 +35,9 @@
   @extend .layout-row;
   min-width: 400px;
 }
+
+.passwordButton {
+  @extend .button-small-filled;
+  justify-content: center;
+  display: flex;
+}
--- a/frontend/src/components/UserInfoCard.tsx
+++ b/frontend/src/components/UserInfoCard.tsx
@@ -93,12 +93,12 @@ export const UserInfoCard = ({ userInfo }: { userInfo: UserInfo }) => {
       }),
     );
   };
-  const passwordOnOk = async (newValue: string) => {
+  const openChangePasswordModal = () => {
     dispatch(
       modalsActions.showModal({
-        modalType: ModalTypes.CONFIRM_PASSWORD_MODAL,
+        modalType: ModalTypes.CHANGE_PASSWORD_MODAL,
         modalProps: {
-          actionData: { id: userInfo.id, password: newValue },
+          id: userInfo.id,
         },
       }),
     );
@@ -150,15 +150,13 @@ export const UserInfoCard = ({ userInfo }: { userInfo: UserInfo }) => {
           </div>
         </div>
         <div>
-          <p className={classes(css.subtitle)}>
-            <Trans i18nKey="Password" />
-          </p>
-          <EditableTextWithButtons
-            onOk={passwordOnOk}
-            value="********"
-            fieldId="password"
-            format=""
-          />
+          <button
+            className={classes(css.passwordButton)}
+            type="button"
+            onClick={() => openChangePasswordModal()}
+          >
+            <Trans i18nKey="Change password" />
+          </button>
         </div>
       </div>
       <ProjectTable items={userInfo.roles} />
--- a/frontend/src/components/forms/FormField.tsx
+++ b/frontend/src/components/forms/FormField.tsx
@@ -8,7 +8,9 @@ import {
 } from 'react';
 import { TextareaAutosize } from '@mui/material';
 import classNames from 'classnames';
+import InfoIcon from '@mui/icons-material/InfoOutlined';
 import css from './FormField.module.scss';
+import { InfoTooltip } from '../InfoTooltip';
 
 const classes = classNames.bind(css);
 
@@ -20,6 +22,7 @@ interface FieldType extends HTMLElement {
 
 export interface FieldProps<T extends FieldType> extends HTMLProps<T> {
   label?: string;
+  labelHint?: string;
   error?: {
     message: string;
     setMessage: (_: string) => void;
@@ -57,6 +60,7 @@ function field<T extends FieldType>(
   return ({
     error = errorState(useState('')),
     label,
+    labelHint,
     innerRef,
     id,
     ...props
@@ -71,7 +75,16 @@ function field<T extends FieldType>(
     }, [ref, error.message, innerRef]);
     return (
       <div className={classes(css.fieldContainer)}>
-        {label && <label htmlFor={id}>{label}</label>}
+        {label && (
+          <label htmlFor={id}>
+            {label}
+            {labelHint && (
+              <InfoTooltip title={labelHint}>
+                <InfoIcon className={classes(css.tooltipIcon)} />
+              </InfoTooltip>
+            )}
+          </label>
+        )}
         <Tag
           {...props}
           ref={ref}
--- a/frontend/src/components/modals/ChangePasswordModal.tsx
+++ b/frontend/src/components/modals/ChangePasswordModal.tsx
@@ -1,34 +1,37 @@
 import { useState, useRef, SyntheticEvent } from 'react';
-import { useHistory } from 'react-router-dom';
 import Alert from '@mui/material/Alert';
 import { useDispatch } from 'react-redux';
 import { Trans, useTranslation } from 'react-i18next';
+import classNames from 'classnames';
 import { ModalTypes, Modal } from './types';
 import { ModalContent } from './modalparts/ModalContent';
 import { ModalFooter } from './modalparts/ModalFooter';
 import { ModalFooterButtonDiv } from './modalparts/ModalFooterButtonDiv';
 import { ModalHeader } from './modalparts/ModalHeader';
 import { errorState, Input } from '../forms/FormField';
-import { paths } from '../../routers/paths';
 import '../../shared.scss';
+import css from './ChangePasswordModal.module.scss';
 
 import { LoadingSpinner } from '../LoadingSpinner';
-
 import { StoreDispatchType } from '../../redux';
 import { userActions } from '../../redux/user';
 
-export const ConfirmPasswordModal: Modal<ModalTypes.CONFIRM_PASSWORD_MODAL> = ({
+const classes = classNames.bind(css);
+
+export const ChangePasswordModal: Modal<ModalTypes.CHANGE_PASSWORD_MODAL> = ({
   closeModal,
-  actionData,
-  deleteUser,
+  id,
 }) => {
-  const history = useHistory();
   const { t } = useTranslation();
   const dispatch = useDispatch<StoreDispatchType>();
-  const [password, setPassword] = useState('');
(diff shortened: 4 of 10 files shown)


PatricKangasmaki:
One issue I found:

image

Now the modal jumps around in size when the error pops, I think it would be a good idea to make the modal size static and the error should just go on two lines if it can't fit. Also, I was wondering if there should be instructions about the password requirements? I think it would be better to have some text like "Password contain atleast 8 characters etc.".

> With this change, that EditableTextWithButtons don't really make sense to me. Should this just be some kind of a button that opens the modal? I temporarily used linkButton for the functionality.  

Never hesitate to contact our designer when in doubt, she's always helpful in things like these 👍

Otherwise this looks great and seems to be working well 👍

DanielMakinen:
> Now the modal jumps around in size when the error pops, I think it would be a good idea to make the modal size static and the error should just go on two lines if it can't fit.

All modals seem to behave this way (at least AddTask and ModifyClient), we might want to make a new card about this if it seems relevant.

> Also, I was wondering if there should be instructions about the password requirements? I think it would be better to have some text like "Password contain atleast 8 characters etc.".

Currently you can see that after you fail to meet the requirements. I agree that it would be good to know before you decide to try some password.

frontend/src/components/UserInfoCard.module.scss
```diff
@@ -35,3 +35,9 @@
   @extend .layout-row;
   min-width: 400px;
 }
+
+.passwordButton {
+  @extend .button-small-filled;
+  justify-content: center;
+  display: flex;
+}
```
juhoeerola, line 43: Please, configure your editor to add newline at the end of the file. That has been the practice on the project, and it also reduces unnecessary noise in git diffs.

Also the indentation is inconsistent in the other css file.

Just formatting the files with prettier should work

frontend/src/components/UserInfoCard.tsx
```diff
@@ -86,27 +86,27 @@ export const UserInfoCard = ({ userInfo }: { userInfo: UserInfo }) => {
   const emailOnOk = async (newValue: string) => {
     dispatch(
       modalsActions.showModal({
-        modalType: ModalTypes.CONFIRM_PASSWORD_MODAL,
+        modalType: ModalTypes.CHANGE_PASSWORD_MODAL,
```
juhoeerola, line 89: The change password modal doesn't make sense here.
Previously the modal was just used for asking the user to confirm the change by providing their password.

A separate modal would probably be good for this, since it feels separate from the functionality of changing a password.

frontend/src/components/UserInfoCard.tsx
```diff
@@ -86,27 +86,27 @@ export const UserInfoCard = ({ userInfo }: { userInfo: UserInfo }) => {
...
+          actionData: { id: userInfo.id },
         },
       }),
     );
   };
   const removeOnClick = () => {
     dispatch(
       modalsActions.showModal({
-        modalType: ModalTypes.CONFIRM_PASSWORD_MODAL,
+        modalType: ModalTypes.CHANGE_PASSWORD_MODAL,
```
juhoeerola, line 109: Same here as `emailOnOk()`

frontend/src/components/UserInfoCard.tsx
```diff
@@ -86,27 +86,27 @@ export const UserInfoCard = ({ userInfo }: { userInfo: UserInfo }) => {
...
         modalProps: {
           actionData: { id: userInfo.id, email: newValue },
         },
       }),
     );
   };
-  const passwordOnOk = async (newValue: string) => {
+  const passwordOnOk = async () => {
```
juhoeerola, line 96: The name `passwordOnOk` doesn't really reflect the current function
maybe `changePassword`, `openChangePasswordModal`, or something like that

Also the `async` is no longer necessary, it was just required by the `EditableTextWithButtons` component
```suggestion
  const changePassword = () => {
```

frontend/src/components/modals/ChangePasswordModal.tsx
```diff
@@ -11,29 +12,35 @@ import { ModalHeader } from './modalparts/ModalHeader';
...
   const formRef = useRef<HTMLFormElement>(null);
 
+  const [errorMessage, setErrorMessage] = useState('');
+  const confirmationError = errorState(useState(''));
+
   const errorHandler = (response?: { data: string; status: number }) => {
     switch (response?.status) {
       case 400: {
-        setErrorMessage(t('Bad request error - is the password correct?'));
+        setErrorMessage(t('Bad request error - is current password correct?'));
```
juhoeerola, line 43: This didn't have a translation to start with, but it would be good to add while changing the key here

frontend/src/components/UserInfoCard.module.scss
DanielMakinen: Now I added automatic linebreaks to vscode, and fixed indentation to 2 spaces.

frontend/src/components/UserInfoCard.tsx
DanielMakinen: This is just me being careless. I didn't notice other functions used that modal and just refactored the word.

I need to be more careful about checking how my changes affect other functions in the future.

frontend/src/components/UserInfoCard.tsx
DanielMakinen: You are right, name changed.

frontend/src/components/modals/ChangePasswordModal.tsx
DanielMakinen: Translation added.

DanielMakinen:
Thanks for the feedback Juho! Changes were made.

frontend/src/components/modals/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+        break;
+      }
+      default:
+        setErrorMessage(t('Internal server error'));
+        break;
+    }
+  };
+
+  const handleConfirm = async () => {
+    if (!formRef.current?.checkValidity()) return;
```
juhoeerola, line 49: Using the submit functionality on the form might be a better for this. This also allows submitting by pressing <kbd>enter</kbd> like other forms. The form will take care of the validation step, but the default action (posting the form) needs to be prevented.

```suggestion
  const handleConfirm = async (e: SyntheticEvent) => {
    e.stopPropagation();
    e.preventDefault();
```

I guess this was pretty much copied from `ConfirmPasswordModal`, so it would be good to change there as well.

form submit 1/3

frontend/src/components/modals/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+    setIsLoading(false);
+    if (action.rejected.match(res)) {
+      errorHandler(res.payload?.response);
+    } else {
+      closeModal(true);
+    }
+  };
+
+  return (
+    <form ref={formRef} className={classes(css.changePasswordForm)}>
```
juhoeerola, line 72: form submit 2/3
```suggestion
    <form onSubmit={handleConfirm} className={classes(css.changePasswordForm)}>
```

frontend/src/components/modals/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </ModalContent>
+      <ModalFooter closeModal={closeModal}>
+        <ModalFooterButtonDiv>
+          {isLoading ? (
+            <LoadingSpinner />
+          ) : (
+            <button
+              className="button-large"
+              type="button"
+              onClick={handleConfirm}
```
juhoeerola, line 132: form submit 3/3
```suggestion
              type="submit"
```

frontend/src/components/modals/types.ts
```diff
@@ -90,6 +91,11 @@ type OwnProps = {
   [ModalTypes.JOIN_PROJECT_MODAL]: { invitation: Invitation };
   [ModalTypes.JOIN_LINK_INVALID_MODAL]: {};
   [ModalTypes.JOIN_LINK_NO_ACCESS_MODAL]: { invitationLink: string };
+  [ModalTypes.CHANGE_PASSWORD_MODAL]: { id: number };
+  [ModalTypes.INFO_MODAL]: { header: string; content: InfoModalContent };
+  [ModalTypes.LEAVE_ROADMAP_MODAL]: {
+    roadmapId: number;
+  };
   [ModalTypes.CONFIRM_PASSWORD_MODAL]:
```
juhoeerola, line 99: Keeping the unnecessary changes to a minimum helps make sense of the actual change.

In this case, moving the "confirm password modal" stuff back above the new "change password modal" lines would help. (same in `ModalRoot`)

frontend/src/components/UserInfoCard.tsx
```diff
@@ -83,22 +83,22 @@ export const UserInfoCard = ({ userInfo }: { userInfo: UserInfo }) => {
...
+        modalType: ModalTypes.CHANGE_PASSWORD_MODAL,
         modalProps: {
-          actionData: { id: userInfo.id, email: newValue },
+          id: userInfo.id,
         },
       }),
     );
   };
-  const passwordOnOk = async (newValue: string) => {
+  const emailOnOk = async (newValue: string) => {
```
juhoeerola, line 96: Here also the reordering should be undone, since `emailOnOk` has not actually changed at all

juhoeerola (changes requested):
There are two things I noticed:
- the form could use `onSubmit` handler on the form, instead of `onClick` on the button
- there is some unnecessary noise from reorderings

Otherwise looks good 👍

frontend/src/components/forms/FormField.tsx
```diff
@@ -69,9 +73,19 @@ function field<T extends FieldType>(
         ref.current.setCustomValidity(error.message);
       }
     }, [ref, error.message, innerRef]);
+    /* eslint-disable */
```
juhoeerola, line 76: Forgot to ask about this. What does this disable?

frontend/src/components/modals/ChangePasswordModal.tsx
DanielMakinen: Changed both ChangePassword and ConfirmPassword modals to use submit.

frontend/src/components/forms/FormField.tsx
DanielMakinen: Nothing anymore, I was using it while testing stuff out. I forgot it there. Thanks for pointing that out.

I've started to double check pushes with difftools, hopefully I won't forget these in the future.

DanielMakinen:
Thanks for the feedback Juho!

I implemented all those changes and I think this pull request might be ready now?

juhoeerola (approved):
Looking good 👍

PR #352, opened 2022-03-24
Fix taskrelation link path

IidaKainu:
The link path was changed a while ago. Perhaps the new path was lost in some conflict


AleksiKytola:
I noticed `tasks.controller.ts` also has a link with /tasks/ still in it, can you add that to this PR too

IidaKainu:
Good catch, added!

AleksiKytola (approved):

PR #357, opened 2022-03-29
Add delete buttons for task ratings

DanielMakinen:
User with Permission.TaskRatingEditOthers (so admin) can delete any ratings from task overview page.

Adds delete-icons to task overview page and upon clicking them it opens a modal to confirm the deletion.

### DIFF
--- a/frontend/src/components/RatingTable.tsx
+++ b/frontend/src/components/RatingTable.tsx
@@ -23,6 +23,7 @@ import { TaskRatingDimension } from '../../../shared/types/customTypes';
 import { BusinessIcon, WorkRoundIcon } from './RoleIcons';
 import colors from '../colors.module.scss';
 import css from './RatingTable.module.scss';
+import { chosenRoadmapIdSelector } from '../redux/roadmaps/selectors';
 
 const classes = classNames.bind(css);
 
@@ -36,6 +37,7 @@ export type RatingRow = FC<{
   style?: CSSProperties;
   user: UserInfo;
   onEdit: (e: MouseEvent) => void;
+  onDelete: () => void;
 }>;
 
 interface RatingTableDef {
@@ -69,6 +71,7 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
   const [listHeight, setListHeight] = useState(0);
   const typeString =
     type === TaskRatingDimension.BusinessValue ? 'value' : 'complexity';
+  const roadmapId = useSelector(chosenRoadmapIdSelector);
 
   useEffect(() => {
     if (!divRef) return;
@@ -96,6 +99,19 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
         modalProps: {
           task,
           edit: true,
+        },
+      }),
+    );
+  };
+
+  const openDeleteModal = (rating: Taskrating) => {
+    if (!roadmapId) return;
+    dispatch(
+      modalsActions.showModal({
+        modalType: ModalTypes.REMOVE_RATING_MODAL,
+        modalProps: {
+          roadmapId,
+          rating,
         },
       }),
     );
@@ -139,6 +155,7 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
             rating={ratings[index]}
             user={userInfo}
             onEdit={openRateModal}
+            onDelete={() => openDeleteModal(ratings[index])}
           />
         )}
       </VariableSizeList>
--- a/frontend/src/components/RatingTableComplexity.tsx
+++ b/frontend/src/components/RatingTableComplexity.tsx
@@ -1,14 +1,19 @@
 import classNames from 'classnames';
 import BuildSharpIcon from '@mui/icons-material/BuildSharp';
-import { useSelector } from 'react-redux';
+import { useSelector, shallowEqual } from 'react-redux';
 import { skipToken } from '@reduxjs/toolkit/query/react';
 import { useTranslation } from 'react-i18next';
 import { chosenRoadmapIdSelector } from '../redux/roadmaps/selectors';
 import { ratingTable, RatingRow } from './RatingTable';
-import { EditButton } from './forms/SvgButton';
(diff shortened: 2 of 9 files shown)


frontend/src/components/RatingTable.tsx
```diff
@@ -101,6 +104,21 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
     );
   };
 
+  const openDeleteModal = (e: MouseEvent, rating: Taskrating) => {
+    e.preventDefault();
+    e.stopPropagation();
+    if (roadmapId === undefined) return;
```
IidaKainu, line 110:
```suggestion
    if (!roadmapId) return;
```

frontend/src/components/RatingTableValue.tsx
```diff
@@ -48,6 +63,9 @@ const TableValueRatingRow: RatingRow = ({ rating, style, user, onEdit }) => {
             user.representativeFor?.some(
               ({ id }) => id === rating.forCustomer,
             ) && <EditButton fontSize="medium" onClick={onEdit} />}
+          {hasEditOthersPermission && (
+            <DeleteButton onClick={(e) => onDelete(e, rating)} />
```
IidaKainu, line 67: This does not necessarily need to be changed, but consider changing the onDelete to `onDelete: () => void;`
```suggestion
            <DeleteButton onClick={onDelete} />
```

The Row's onDelete would turn to `onDelete={openDeleteModal(ratings[index])}`

I suggest this, since I don't see why the event default functionality or propagation would need to be cancelled in this case and the rating can be accessed in the RatingTable.

frontend/src/components/RatingTableComplexity.tsx
```diff
@@ -40,6 +51,13 @@ const TableComplexityRatingRow: RatingRow = ({
           {createdBy?.email ?? `<${t('deleted account')}>`}
         </div>
         <div className={classes(css.rightSide)}>
+          {hasEditOthersPermission && (
+            <DeleteButton
+              onClick={(e) => {
+                onDelete(e, rating);
+              }}
+            />
```
IidaKainu, line 59:
```suggestion
            <DeleteButton onClick={(e) => onDelete(e, rating)} />
```

IidaKainu (changes requested):
Nice that you added a confirmation modal for this 👍 
The modal text could be more descriptive though, so the user can confirm they are deleting the right rating. Maybe something like the following

"Are you sure you want to remove rating by {{userEmail}} for {{customerName}}"

frontend/src/components/RatingTable.tsx
DanielMakinen: Good catch

frontend/src/components/RatingTableValue.tsx
DanielMakinen: > The Row's onDelete would turn to onDelete={openDeleteModal(ratings[index])}

Everything else worked fine, but with this line it seemed that I needed to wrap that function.

`onDelete={() => openDeleteModal(ratings[index])}`

Otherwise, it will complain about "Type 'void' is not assignable to type '() => void'.ts(2322)"

Not sure whats that about.

DanielMakinen:
Thanks again Iida for the feedback!

I modified the warning to be more informative. I needed to get the email of the rater and the name of their company by querying the API in the modal file.

```
  const { data: rater } = apiV2.useGetRoadmapUsersQuery(
    roadmapId ?? skipToken,
    selectById(rating.createdByUser),
  );
  const { data: customer } = apiV2.useGetCustomersQuery(
    roadmapId ?? skipToken,
    selectById(rating.forCustomer),
```

Another way would have been to pass the email and company name from RatingTableValue/complexity files, where these things are already queried. These could be passed through onDelete-functions, which would add more bloated code for the other files. Any opinions which is the better way to do this?

frontend/src/components/RatingTableValue.tsx
DanielMakinen: Right, cause without the wrapping, its trying to pass the returned value of `openDeleteModal(ratings[index])` instead of the function itself.

IidaKainu: You are correct! I'm sorry for any extra work caused by the bad example

DanielMakinen: All good 👍

Made me read couple of articles about function passing in React and now I got better understanding of whats going on.

IidaKainu:
The code is looking good! Small fix for developers added - they do not have a customer so the modal was empty

> Another way would have been to pass the email and company name from RatingTableValue/complexity files, where these things are already queried. These could be passed through onDelete-functions, which would add more bloated code for the other files. Any opinions which is the better way to do this?

Good question. In terms of readability or reusability (not as relevant in this case) it would be better to let the modal handle the querying.  Furthermore, it is unlikely a project would have a huge amount of users and customers, so re-fetching the data from Redux is not a burden in this case.

frontend/src/components/modals/RemoveRatingModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </ModalHeader>
+      <ModalContent>
+        <div className="modalCancelContent">
+          <AlertIcon />
+          <div>
+            {customer ? (
+              <Trans
+                i18nKey="Remove representative rating warning"
+                values={{
+                  rater: rater?.email ?? 'deleted user',
```
juhoeerola, line 65:
```suggestion
                  rater: rater?.email ?? t('deleted account'),
```

It would be good to use the same (translated) default for deleted users as elsewhere.  (also for developer ratings)

juhoeerola (changes requested):

frontend/src/components/modals/RemoveRatingModal.tsx
IidaKainu: 👍 Fixed

juhoeerola (approved):

IidaKainu (approved):

DanielMakinen:
Thanks for the fixes!

PR #362, opened 2022-03-31
Fix RateTaskModal loading issue.

PatricKangasmaki:
Rate task modal tried to apply task ratings to the businessValueRatings state before fetching task correctly. This only happened when you tried giving ratings from the task modal that opens automatically after creating a new task.

I changed the parameters from TaskId to Task, this way the task can't be either loading or undefined.

### DIFF
--- a/frontend/src/components/modals/AddTaskModal.tsx
+++ b/frontend/src/components/modals/AddTaskModal.tsx
@@ -58,7 +58,7 @@ export const AddTaskModal: Modal<ModalTypes.ADD_TASK_MODAL> = ({
       };
 
       try {
-        const res = await addTaskTrigger({
+        const task = await addTaskTrigger({
           roadmapId: chosenRoadmapId!,
           task: req,
         }).unwrap();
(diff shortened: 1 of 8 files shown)


frontend/src/components/modals/AddTaskModal.tsx
```diff
@@ -58,7 +58,7 @@ export const AddTaskModal: Modal<ModalTypes.ADD_TASK_MODAL> = ({
       };
 
       try {
-        const res = await addTaskTrigger({
+        const task = await addTaskTrigger({
           roadmapId: chosenRoadmapId!,
           task: req,
         }).unwrap();
```
PatricKangasmaki, line 64: I noticed that the addTaskTrigger returns a task with no property `ratings` at all when the request has no ratings (or the ratings is an empty array) since server returns `task: ..., ratings: undefined` in that case. Should this be looked at deeper, or this the `task: { ...task, ratings: [] },` fix enough for now? Currently RateTaskModal can't handle a task with no ratings property.

frontend/src/components/modals/AddTaskModal.tsx
```diff
@@ -68,7 +68,7 @@ export const AddTaskModal: Modal<ModalTypes.ADD_TASK_MODAL> = ({
           modalsActions.showModal({
             modalType: ModalTypes.RATE_TASK_MODAL,
             modalProps: {
-              taskId: res.id,
+              task: { ...task, ratings: [] },
               edit: false,
             },
           }),
```
PatricKangasmaki, line 74: The current fix I mentioned above is located here

AleksiKytola (approved):

PR #363, opened 2022-03-31
Query fixes

AleksiKytola:
Skip version queries if logged in user does not have permssion to query versions
(this used to create a failed query with status 403 every time certain components rendered on non-admins)

Skip roadmapsidebar roadmaps query if there is no logged in user
(this used to spam failed queries on login page)


IidaKainu (approved):
Seems to work as intended
I had not noticed the failed query spam in the login page before - great fix 🔥

PR #364, opened 2022-04-04
Make selectedRoadmapId nullable, set to null on 404

AleksiKytola:
https://trello.com/c/kbrJoN6F/520-fix-404-sivun-tyylit-kun-roadmap-on-auki

Made roadmapId nullable, also made 404 page deselect roadmap in order to hide sidebar as the trello card proposed. Returning to a previously valid roadmap still works with the browsers back button.

### DIFF
--- a/frontend/src/components/NavBar.tsx
+++ b/frontend/src/components/NavBar.tsx
@@ -36,7 +36,7 @@ export const NavBar = () => {
           <VisdomLogo />
         </div>
       )}
-      {pathname.startsWith(paths.userInfo) && roadmapId === undefined && (
+      {pathname.startsWith(paths.userInfo) && !roadmapId && (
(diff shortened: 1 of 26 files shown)


IidaKainu (approved):
Functions great 👍 

First I wondered why set the roadmap to null, when undefined could be (technically) as acceptable. But I suppose there is a difference semantically - in 404 selectedRoadmapId has no value, whereas in overview it has not been assigned yet.

AleksiKytola:
> First I wondered why set the roadmap to null, when undefined could be (technically) as acceptable. But I suppose there is a difference semantically - in 404 selectedRoadmapId has no value, whereas in overview it has not been assigned yet.

Setting to undefined would probably be functionally equal but as a javascript coding practice its the most correct to use `null` as an intentionally blank value so I went and properly changed that also

PR #365, opened 2022-04-06
Add format for multiline text in EditableText

IidaKainu:
For example, "format: 'multiline'" can be passed to data in overviews.

https://trello.com/c/wMOi6X6N/598-taskoverview-description-render%C3%B6i-newlinet


juhoeerola (approved):
👍

PR #367, opened 2022-04-07
Clear taskrating

IidaKainu:
Add the possibility to clear ratings in the taskrating modal by re-clicking the current rating
  - If the modal is in edit-mode, the cleared rating will be deleted
  - Otherwise, the cleared rating is not sent on submit

Taskrating deletion is added to the patch-route to handle everything in one transaction
  - If the rating value is zero, it is deleted in the patch-route
  - PatchTaskratings is renamed to updateRaskratings

### DIFF
--- a/frontend/src/components/modals/RateTaskModal.tsx
+++ b/frontend/src/components/modals/RateTaskModal.tsx
@@ -76,7 +76,10 @@ export const RateTaskModal: Modal<ModalTypes.RATE_TASK_MODAL> = ({
   );
   const roadmapId = useSelector(chosenRoadmapIdSelector);
 
-  const [patchTaskratings, patchStatus] = apiV2.usePatchTaskratingsMutation();
+  const [
+    updateTaskratings,
+    updateStatus,
+  ] = apiV2.useUpdateTaskratingsMutation();
   const [addTaskratings, addStatus] = apiV2.useAddTaskratingsMutation();
 
   const dimension =
@@ -146,7 +149,7 @@ export const RateTaskModal: Modal<ModalTypes.RATE_TASK_MODAL> = ({
     event.stopPropagation();
 
     const changed = businessValueRatings.filter((rating) => rating.changed);
-    const action = edit ? patchTaskratings : addTaskratings;
+    const action = edit ? updateTaskratings : addTaskratings;
     if (!roadmapId || changed.length === 0) return;
     try {
       await action({ roadmapId, ratings: changed, taskId: task.id }).unwrap();
@@ -233,7 +236,7 @@ export const RateTaskModal: Modal<ModalTypes.RATE_TASK_MODAL> = ({
       </ModalContent>
       <ModalFooter closeModal={closeModal}>
         <ModalFooterButtonDiv>
-          {patchStatus.isLoading || addStatus.isLoading ? (
+          {updateStatus.isLoading || addStatus.isLoading ? (
             <LoadingSpinner />
           ) : (
             <button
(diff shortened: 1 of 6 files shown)


server/src/api/taskratings/taskratings.controller.ts
```diff
@@ -186,13 +186,23 @@ export const patchTaskratings: RouteHandlerFnc = async (ctx) => {
...
-    Promise.all(
+  const result = await Taskrating.transaction(async (trx) => {
+    const toDelete = foundRatings.filter(({ value }) => !value);
+    const deleted = await Promise.all(
+      toDelete.flatMap(({ rating }) =>
+        rating ? Taskrating.query(trx).findById(rating.id).delete() : [],
+      ),
+    );
+
+    if (deleted.length !== toDelete.length) throw new Error();
```
IidaKainu, line 197: Just want to highlight this, as it might be questionable. I wanted to rollback the transaction in case some of the deletions are not made, hence the error. I am not sure though if it is required and whether to add an error description (string)

frontend/src/components/modals/RateTaskModal.tsx
```diff
@@ -138,28 +140,20 @@ export const RateTaskModal: Modal<ModalTypes.RATE_TASK_MODAL> = ({
         !rating.customer || rating.customer.roadmapId === task?.roadmapId,
     )
     // show only new or existing ratings depending on edit value
-    .filter(({ value }) => (edit ? value : !value));
+    .filter(({ value }) => (edit ? value : !value))
+    .map((rating) => ({
+      ...rating,
+      ...(!rating.comment && { comment: undefined }),
+    }));
```
juhoeerola, line 147: When editing a rating this will make a distinction between untouched rating with no comment (comment is `undefined`) and a rating where empty comment is first edited but then cleared (comment is the empty string).

This shows in the submit button being enabled in the latter case, even though no meaningful change has been made.

So, unless I'm missing something, I think this should be removed.

server/src/api/taskratings/taskratings.controller.ts
juhoeerola: I suppose this could happen if more than one user remove the same rating at the same time.
I don't think that should be necessarily be an error, since the deletion failed because the item was already gone.

server/src/api/taskratings/taskratings.controller.ts
```diff
@@ -186,13 +186,23 @@ export const patchTaskratings: RouteHandlerFnc = async (ctx) => {
...
 
-  const updated = await Taskrating.transaction((trx) =>
-    Promise.all(
+  const result = await Taskrating.transaction(async (trx) => {
+    const toDelete = foundRatings.filter(({ value }) => !value);
+    const deleted = await Promise.all(
+      toDelete.flatMap(({ rating }) =>
+        rating ? Taskrating.query(trx).findById(rating.id).delete() : [],
+      ),
+    );
```
juhoeerola, line 195:
```suggestion
   const toDelete = foundRatings.flatMap(({ value, rating }) =>
     !value && rating ? [rating.id] : [],
   );
   const deleted = await Taskrating.query(trx).findByIds(toDelete).delete();
```

This would be better to do the other way around: a single query with multiple ids.

juhoeerola (changes requested):

frontend/src/components/modals/RateTaskModal.tsx
IidaKainu: I think it related to the logic in detecting changed edited ratings. Improved the predicate so it is not needed anymore.

IidaKainu:
Great points, thanks Juho. I made changes according to the feedback.

juhoeerola (approved):

PR #368, opened 2022-04-07
Fix 404-routing of reload of some overview pages

IidaKainu:
Client and task overview pages routed to 404 on reload. This was due to delay in fetching the chosenRoadmapSelector.

To fix the problem, the page components can show a waiting indicator while the roadmapId is undefined or null. This is possible, since RoamapRouter handles setting the chosenRoadmapId and routing to 404 if it does not exist.


juhoeerola (approved):
👍

PR #370, opened 2022-04-08
Widen milestone columns

IidaKainu:
The milestone width is increased and task list paddings are decreased.


frontend/src/components/SortableTaskList.module.scss
```diff
@@ -18,7 +18,7 @@
   overflow-y: auto;
   min-height: 100px;
   height: inherit;
-  padding: 16px;
+  padding: 1px;
   padding-top: 0px;
   scrollbar-width: thin;
```
juhoeerola, line 24:
```scss
.taskName {
  @extend .taskNameContainer;
  width: 175px;
}
```
This limits the task name to narrower than it would have space for. It's applied with `.leftSideDiv` from SortableTask.module.scss, where also the width should be adjusted (or maybe just extend `.taskNameContainer`, instead of this `.taskName`)

```scss
.leftSideDiv {
  @extend .taskName;
}
```

juhoeerola (changes requested):

IidaKainu:
Good points. Removed the taskName-class definition and widened search field's padding according to .shadow-input's spread-radius.

frontend/src/components/SortableTaskList.module.scss
```diff
@@ -29,19 +29,14 @@
...
-.taskName {
-  @extend .taskNameContainer;
-  width: 175px;
-}
-
 .measureTaskName {
-  @extend .taskName;
+  @extend .taskNameContainer;
   visibility: hidden;
 }
```
juhoeerola, line 41: Just removing this unfortunately breaks the measuring. 
The measure div should *always* have the *exactly same width* as what the name will have in the real list item. Otherwise there will be cases where the measured text uses different number of rows than the actual text.

These available widths are different for the milestones and for the unordered-list, 208.5 and 175.75 pixels respectively as I checked. Both happen to be narrower than the 210 pixels that is given for the measuring div.

I think the measuring is overall bit too fragile at the moment, breaking too easily on style changes and the breaking being too easy to not notice.

I think using hard coded widths and verifying that those have actually the full width in the dom, is the most reasonable without larger refactoring. Although a better way to do this would be great...

juhoeerola (changes requested):

IidaKainu:
The same row component is now used to measure name div, so the div width is consistent with the list.

IidaKainu:
Fix the bug where the toggling of ExpandableColumn would break measuring logic.

frontend/src/components/ExpandableColumn.tsx
```diff
@@ -12,26 +12,38 @@ export const ExpandableColumn: FC<{
...
-        [css.minimized]: !expanded,
-      })}
-    >
+}> = ({ expanded, onToggle, title, children, className, style }) => {
+  // delay showing children, to avoid breaking possible measuring divs
+  const [showChildren, setShowChildren] = useState(expanded);
+
+  useEffect(() => {
+    if (!expanded) setShowChildren(false);
+    else setTimeout(() => setShowChildren(true), 300);
```
juhoeerola, line 21:
```suggestion
    if (!expanded) {
      setShowChildren(false);
    } else {
      const t = setTimeout(() => setShowChildren(true), 300);
      return () => clearTimeout(t);
    }
```

If toggling quickly expanded -> closed, the timeout will stay alive and set the children to be rendered when the column is already closed.

Also it would be good to cancel the timeout. I'm not 100% sure if this on it's own is enough for the above bug though.

juhoeerola (changes requested):

juhoeerola (approved):
Seems to work well 👍

IidaKainu:
Added ugly brackets to if-oneliner as wished - merging.

PR #371, opened 2022-04-08
Status mapping for integrations

juhoeerola:
Added required methods to integration interface:
- `columns(boardId)` for getting the columns (lists in trello terminology) on the specified board
- optional `columnId` property in `ImportedTask`
- made `status` optional in `ImportedTask`
- the status is set from the first that applies:
   - status mapping based on `columnId`
   - `status` from `ImportedTask`
   - `TaskStatus.NOT_STARTED`

Added required api routes and methods for frontend / server.

Added functional but not too nice settings page.

Moved personal auth-token settings to user info page, as that is per user, not per roadmap.

---

In Jira the imported task includes a "status", which in turn may be mapped to some column (column may have multiple statuses mapped to it, but a status can only be mapped to one column). This status id is used for the "columns". The name is "column-name (status-name)" or "column-name" if column and status both have the same name.


frontend/src/api/api.ts
```diff
@@ -285,15 +285,67 @@ export const apiV2 = createApi({
...
+      }),
+    }),
+    setIntegrationStatusMapping: build.mutation<
+      void,
+      {
+        id: number;
+        name: string;
+        roadmapId: number;
+        fromColumn: string;
+        toStatus: number;
```
IidaKainu, line 341: Would the type TaskStatus be more specific here?
```suggestion
        toStatus: TaskStatus;
```

IidaKainu (changes requested):
I do not feel like the best person to review the backend controllers related to the integrations, but the changes look good to me on the surface level. The frontend for configuring a integration is a bit confusing, but functional (which is really the goal in this pr, so well done 👍)

Found a small suggestion and the styles in `IntegrationConfiguration.tsx` could be moved to the scss-file

IidaKainu:
Almost forgot to ask what is the personal auth token for in the user page? I don't think I needed it while testing the trello-configuration.

juhoeerola:
> Almost forgot to ask what is the personal auth token for in the user page?

It's totally unrelated to the integrations, it can be used instead of cookies to authenticate to the roadmapper to use the apis.

frontend/src/api/api.ts
juhoeerola: Yes, it would, good point

IidaKainu (approved):
Looks good to me
Feel free to request review from someone else too if you want a second opinion on the backend

PR #372, opened 2022-04-08
Disable MilestonesEditor features while loading

IidaKainu:
Any buttons that open a modal are disabled while versions are patched in MilestonesEditor.
- Loading cursor is applied to the whole milestone and expandable task list

### DIFF
--- a/frontend/src/components/forms/SvgButton.module.scss
+++ b/frontend/src/components/forms/SvgButton.module.scss
@@ -1,17 +1,27 @@
 @import '../../shared.scss';
 
 .svgButtonIcon {
+  &:not(.disabled) {
+    &:hover path {
+      fill: var(--icon-hover-color, $COLOR_BLACK60);
+    }
+    cursor: pointer;
+  }
+
   path {
     fill: var(--icon--color, $COLOR_BLACK40);
     transition: 150ms ease-out;
   }
-  &:hover path {
-    fill: var(--icon-hover-color, $COLOR_BLACK60);
+  &.disabled path {
+    fill: var(--icon-disabled-color, $COLOR_BLACK40);
   }
-  cursor: pointer;
 }
 
 .closeButton {
   width: 24px !important;
   height: 24px !important;
+}
+
+a {
+  line-height: 1;
 }
--- a/frontend/src/components/forms/SvgButton.tsx
+++ b/frontend/src/components/forms/SvgButton.tsx
@@ -17,6 +17,7 @@ const classes = classNames.bind(css);
 
 type ButtonProps = {
   onClick?: (event: MouseEvent) => void;
+  disabled?: boolean;
 };
 
 type ExtraProps = {
@@ -36,16 +37,19 @@ function svgButton<E = {}>(
     ...defaults
   }: ExtraProps & { [K in keyof E]?: E[K] },
 ): SvgButton<Pick<E, Exclude<keyof E, keyof ExtraProps>>> {
-  return ({ href, ...props }) => {
+  return ({ href, disabled, onClick, ...props }) => {
     const icon = (
       <Icon
-        className={classes(css.svgButtonIcon, className)}
+        className={classes(css.svgButtonIcon, className, {
+          [css.disabled]: disabled,
+        })}
         style={{ '--icon--color': iconColor, '--icon-hover-color': hoverColor }}
+        onClick={disabled ? undefined : onClick}
         {...defaults}
         {...props}
       />
     );
-    return href ? <a href={href}>{icon}</a> : icon;
+    return disabled || !href ? icon : <a href={href}>{icon}</a>;
   };
 }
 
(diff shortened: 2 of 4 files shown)


frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -322,21 +324,36 @@ export const MilestonesEditor = () => {
...
+                      })}
+                    >
+                      {disableDrag ? (
+                        <>
+                          <DeleteIcon /> <SettingsSharpIcon />
+                        </>
+                      ) : (
+                        <>
+                          <DeleteButton
+                            onClick={deleteVersionClicked(version.id)}
```
juhoeerola, line 339:
```suggestion
                          <DeleteButton
                            disabled={disableDrag}
                            onClick={deleteVersionClicked(version.id)}
```

I'd rather add `disabled` property to the svg button, like normal buttons have. That way it can be used for all those buttons.

The disabled svg button would not use the link even if href was provided, and hover & pointer styles would not be applied.

To keep the height the same with or without a link, the `a` tag could be given `line-height: 1`

juhoeerola (changes requested):

frontend/src/pages/MilestonesEditor.module.scss
```diff
@@ -45,6 +45,18 @@
...
+      width: 21px;
+      height: 23px;
+      position: relative;
+      top: 1px;
+
+      path {
+        fill: $COLOR_BLACK40;
+      }
+    }
+  }
```
juhoeerola, line 59: This should be unnecessary, if the styles are applied on the svg button disabled styles

frontend/src/components/forms/SvgButton.module.scss
```diff
@@ -11,6 +11,17 @@
...
+  line-height: 1;
+  .svgButtonIcon {
+    cursor: default;
+
+    path {
+      fill: var(--icon-disabled-color, $COLOR_BLACK40);
+    }
+  }
+}
+
```
juhoeerola, line 24:
```ts
  &:not(.disabled) {
    &:hover path {
      fill: var(--icon-hover-color, $COLOR_BLACK60);
    }
    cursor: pointer;
  }
```

It might be better to just not apply these in `.svgButtonIcon` when disabled, then e.g. the loading cursor would still show when hovering the icons.

(then the `.disabled` class should be added to the icon obviously)

frontend/src/components/forms/SvgButton.tsx
```diff
@@ -36,15 +37,17 @@ function svgButton<E = {}>(
...
       <Icon
         className={classes(css.svgButtonIcon, className)}
         style={{ '--icon--color': iconColor, '--icon-hover-color': hoverColor }}
+        onClick={disabled ? undefined : onClick}
         {...defaults}
         {...props}
       />
     );
+    if (disabled) return <div className={classes(css.disabled)}>{icon}</div>;
     return href ? <a href={href}>{icon}</a> : icon;
```
juhoeerola, line 51:
```suggestion
    return href && !disabled ? <a style={{ lineHeight: 1 }} href={href}>{icon}</a> : icon;
```

The wrapping div should not be needed, or am I missing something?

Also, I'd say that `line-height: 1` for the anchor is better than the `height: width + 2 * 1px; top: 1px;` dance.
Then the width / height should stay the same if anchor or not.

juhoeerola (changes requested):
The disabled icon styling could be done exclusively in the `SvgButton` based on the `disabled` prop

frontend/src/components/forms/SvgButton.tsx
IidaKainu: You are right - I was probably confused by MilestonesEditor's forgotten svg styles.

PatricKangasmaki:
Found a few things, otherwise this is looking great.

The cursor currently shows `default` when hovering over delete and settings icons while loading. It could be changed to loading as well. Same thing with the Add new milestone button.

I also noticed that you can change page and come back to milestones editor and then you can move things around and cursor doesn't indicate loading even though it's still loading in the back. That can cause situations where you have same task in multiple milestones (reloading doesn't fix it but moving tasks in milestones will). Seems like that it even counts towards the task values in every milestones where it is.

image

IidaKainu:
Thanks for the review Patric!

> The cursor currently shows default when hovering over delete and settings icons while loading

This is already fixed

> Same thing with the Add new milestone button

Added a fix for this

> I also noticed that you can change page and come back to milestones editor and then you can move things around and cursor doesn't indicate loading even though it's still loading in the back. That can cause situations where you have same task in multiple milestones (reloading doesn't fix it but moving tasks in milestones will). Seems like that it even counts towards the task values in every milestones where it is.

I'll add a card about this to Trello, since the problem is applicable to task map as well and should be done in a separate pr.

juhoeerola (changes requested):
I agree, the navigating away and back can be dealt in a separate issue. I couldn't reproduce tasks duplicating, but some weirdness there was for sure.

Other than that looks good 👍 

Just remove the undefined (unless I'm mistaken) css-class

IidaKainu:
> Just remove the undefined (unless I'm mistaken) css-class

Removed. Should be finally in merge-condition

juhoeerola (approved):

PR #373, opened 2022-04-08
Fix SortableTaskList drag bug on search

juhoeerola:
The dragged task was taken from the full list, not the searched one.


IidaKainu (approved):
That was an obscure one. Well caught 🔥

PR #375, opened 2022-04-11
Change unstaged tasks -list to open automatically when taskmap canvas…

patricvincit:
This change opens unstaged tasks -list automatically when none of the tasks have been placed on the task map yet. Added it for improved usability since that's what you'd do first if the taskmap was empty


juhoeerola (changes requested):

PatricKangasmaki:
Great point, I completely missed the fact that `stagedTasks` could be used straight away without any filtering. Added the suggestion 👍

juhoeerola (approved):

PR #379, opened 2022-04-13
Completed task can no longer be considered unrated

juhoeerola:


IidaKainu (commented):
I am wondering should we consider task completion in hasRatingsOnEachDimension-function as well? A completed task without either no value or work rating is not shown in MilestonesEditor.

juhoeerola:
I suppose completed task can't be "unrated" in the meaning "needs to be rated", but it doesn't mean the task would actually have ratings. I think the status should not be the concern of `hasRatingsOnEachDimension`, but to be checked separately when we want "ratings on each dimension *or* task is completed".

IidaKainu (approved):
>  I think the status should not be the concern of hasRatingsOnEachDimension, but to be checked separately when we want "ratings on each dimension or task is completed".

Good point

PR #380, opened 2022-04-14
Add topbar with visualizations to milestone planner

AleksiKytola:
Refactored PercentageBar to its own component, adjusted interface of components for more flexibility (stakes visualization accepts an array of versions instead of just one)

Relocated the info icon to the topbar, it was looking weirdly out of place with the original absolute positioning, and I felt it was easier to notice up in the topbar

Currently the stakes visualization calculates the stakes for all roadmaps milestones, milestone selection is in another card and this component shall be updated then to only use the selected milestones. 

image

### DIFF
--- a/frontend/src/components/CustomerWeightsVisualization.module.scss
+++ b/frontend/src/components/CustomerWeightsVisualization.module.scss
@@ -27,3 +27,7 @@
     color: $COLOR_AZURE;
   }
 }
+
+.stakes.light {
+  filter: opacity(70%);
+}
--- a/frontend/src/components/CustomerWeightsVisualization.tsx
+++ b/frontend/src/components/CustomerWeightsVisualization.tsx
@@ -1,16 +1,20 @@
-import { FC } from 'react';
+import { FC, useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+import { skipToken } from '@reduxjs/toolkit/query/react';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 import Tooltip from '@mui/material/Tooltip';
+import { chosenRoadmapIdSelector } from '../redux/roadmaps/selectors';
+import { CustomerStakes } from '../redux/roadmaps/types';
+import css from './CustomerWeightsVisualization.module.scss';
+import { apiV2 } from '../api/api';
 import { Dot } from './Dot';
 import { percent } from '../utils/string';
-import { CustomerStakes } from '../redux/roadmaps/types';
-import css from './CustomerStakesVisualization.module.scss';
 import { PercentageBar } from './PercentageBar';
 
 const classes = classNames.bind(css);
 
-export const StakesTooltipContent: FC<{
+export const CustomerWeightsTooltipContent: FC<{
   customerStakes: CustomerStakes[];
   totalValue: number;
   showPercentageBar?: boolean;
@@ -19,11 +23,15 @@ export const StakesTooltipContent: FC<{
   const { t } = useTranslation();
 
   if (customerStakes.length === 0)
-    return <div className={classes(css.stakesTooltip)}>{t('No ratings')}</div>;
+    return (
+      <div className={classes(css.stakesTooltip)}>
+        {t('No client weights set')}
+      </div>
+    );
 
   return (
     <div className={classes(css.stakesTooltip)}>
-      {!noTitle && <div>{t('Client shares')}</div>}
+      {!noTitle && <div>{t('Client weights')}</div>}
       {showPercentageBar && (
         <PercentageBar
           stakes={customerStakes}
@@ -49,7 +57,7 @@ export const StakesTooltipContent: FC<{
   );
 };
 
-export const CustomerStakesVisualization: FC<{
+const CustomerWeightsPercentageBar: FC<{
   customerStakes: CustomerStakes[];
   totalValue: number;
   vertical?: boolean;
@@ -72,7 +80,7 @@ export const CustomerStakesVisualization: FC<{
         tooltip: classes(css.tooltip),
       }}
       title={
-        <StakesTooltipContent
+        <CustomerWeightsTooltipContent
           customerStakes={customerStakes}
           totalValue={totalValue}
         />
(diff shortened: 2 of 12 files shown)


AleksiKytola:
Also added a commit to make planner milestones wider

frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -257,6 +259,51 @@ export const MilestonesEditor = () => {
...
+        </div>
+        <div className={classes(css.topbarShares)}>
+          <Trans i18nKey="Target weighted shares" />
+          {roadmapsVersions && (
+            <CustomerWeightsVisualization
+              width={250}
+              height={30}
+              barWidth={30}
+            />
+          )}
```
PatricKangasmaki, line 301: The design seems a bit off compared to figma. I'd make the height and width of the bars a little smaller, for example 15px-20px and 200px, as they currently take away the main focus on the page.

```suggestion
          {roadmapsVersions && (
            <TaskValueCreatedVisualization
              versions={[...roadmapsVersions]}
              width={200}
              height={20}
              barWidth={15}
            />
          )}
        </div>
        <div className={classes(css.topbarShares)}>
          <Trans i18nKey="Target weighted shares" />
          {roadmapsVersions && (
            <CustomerWeightsVisualization
              width={200}
              height={20}
              barWidth={15}
            />
          )}
```

frontend/src/pages/MilestonesEditor.module.scss
```diff
@@ -102,3 +104,20 @@
...
+  padding: 10px 0px;
+}
+
+.topbarShares {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  margin: 0px 10px;
+}
```
PatricKangasmaki, line 123: Titles could have the pre-title font here or you can add class to the div's of the titles. Also some gap between the elements look good.

```suggestion
.topbarShares {
  @extend .typography-pre-title;
  display: flex;
  flex-direction: column;
  align-items: center;
  justify-content: center;
  gap: 5px;
  margin: 0px 10px;
}
```

frontend/src/components/PercentageBar.module.scss
```diff
@@ -0,0 +1,16 @@
...
+  // horizontal
+  width: 100%;
+  height: max-content;
+  flex-direction: row;
+
+  &.vertical {
+    height: 100%;
+    width: max-content;
+    flex-direction: column;
+  }
```
PatricKangasmaki, line 15: I'd change the `horizontal` to it's own class just like the vertical one for consistency. Or if you want to keep horizontal as the default you could just change it to:

```
@import '../shared.scss';

.stakes {
  @extend .layout-row;
  border-radius: 10px;
  overflow: hidden;
  width: 100%;
  height: max-content;

  &.vertical {
    height: 100%;
    width: max-content;
    flex-direction: column;
  }
}
```

frontend/src/components/TaskValueCreatedVisualization.module.scss
```diff
@@ -5,7 +5,9 @@
   @extend .layout-row;
   gap: 47px;
   width: 100%;
+  height: 100%;
   .stakesDescription {
     width: 100%;
+    height: 100%;
```
PatricKangasmaki, line 11: Is there a reason to define width and height twice here with the same values?

PatricKangasmaki (changes requested):
Looking good 👍 
Added a few suggestions below. Was there some technical reason not to display percentages right away in the percentage bars (figma design) or was it just better looking with tooltips? It might be a bit faster to see the actual differences between actual and target values if you could see the % values right away without having to hover.

PatricKangasmaki:
Few more things that I found:
This link to client weight on the right could also be added in this PR and target weight colors seem to be a little lighter (which is good so it can be easily distinguished from the actual weights)

image

frontend/src/pages/MilestonesEditor.module.scss
PatricKangasmaki: `display: flex` and `flex-direction: column;` can be changed to `@extend .layout-column;` here as well

frontend/src/components/TaskValueCreatedVisualization.module.scss
AleksiKytola: Seems like the whole .stakesDescription class is unnecessary

AleksiKytola:
Applied recommended style fixes, added the button to navigate to client weights next to the target weighted shares bar.
image

frontend/src/components/CustomerWeightsVisualization.tsx
```diff
@@ -0,0 +1,162 @@
...
+          color,
+        })),
+    );
+  }, [customers, setData]);
+
+  return (
+    <div className={classes(css.stakes)}>
+      <div style={{ width, height }}>
+        <div className={classes(css.customerBarWithLink)}>
+          <CustomerWeightsPercentageBar
```
IidaKainu, line 136: Could CustomerStakesVisualization be used here instead of CustomerWeightsPercentageBar? It looks like the components are the same unless I missed something

frontend/src/components/PercentageBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+        stakes
+          .filter(({ value }) => value > 0)
+          .map((entry) => (
+            <BarSection
+              key={entry.id}
+              size={entry.value / totalValue}
+              color={entry.color}
+              vertical={vertical}
+              width={width}
+            />
```
IidaKainu, line 50: An option to show the percentages in the bar could be added here. Though it requires a bit of "prop drilling" for showPercentages
```suggestion
            >
              {showPercentages && percent(0).format(entry.value / totalValue)}
            </BarSection>
```

IidaKainu (changes requested):
Added a few suggestions, otherwise looks good to me 👍

juhoeerola (changes requested):

frontend/src/components/CustomerWeightsVisualization.tsx
AleksiKytola: Currently its just slightly different tooltips

frontend/src/components/PercentageBar.tsx
AleksiKytola: I preferred not to add it yet since I dont have a solution to the problem where the section is too small to have a readable percentage

frontend/src/components/TaskValueCreatedVisualization.tsx
```diff
@@ -20,20 +23,36 @@ interface VersionComplexityAndTotalValue extends Version {
...
+  vertical?: boolean;
+}> = ({ versions, width, height, noTooltip, vertical, barWidth = 37 }) => {
   const roadmapId = useSelector(chosenRoadmapIdSelector);
   const { data: customers } = apiV2.useGetCustomersQuery(
     roadmapId ?? skipToken,
   );
+
   const [data, setData] = useState<CustomerStakes[]>([]);
   const totalValue = data.reduce((acc, { value }) => acc + value, 0);
+  const versionsWithValueAndComplexity: VersionComplexityAndTotalValue[] = versions.map(
```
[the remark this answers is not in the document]
AleksiKytola: Weird that I didnt notice infinite updates going on, performance was fine. But anyway it does seem unnecessary. I just noticed the original component was passed a VersionComplexityAndTotalValue so when I changed it to take multiple versions I kept it

frontend/src/components/PercentageBar.tsx
IidaKainu: I see - that's understandable

AleksiKytola:
Applied Iida's CSS refactors and Juho's fixes :+1:

IidaKainu (approved):

frontend/src/components/CustomerWeightsVisualization.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  @extend .layout-row;
+  text-align: center;
+  align-items: center;
+  justify-content: center;
+
+  a {
+    position: absolute;
+    right: -14.5em;
+    top: 0.4em;
+  }
```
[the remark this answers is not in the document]
AleksiKytola: Added

frontend/src/components/PercentageBar.module.scss
```diff
@@ -0,0 +1,32 @@
...
+
+.barSection {
+  z-index: 10;
+  user-select: none;
+  background-color: var(--color);
+
+  height: var(--barThickness);
+  width: calc(100% * var(--size, 1));
+  &.horizontal {
+  }
```
PatricKangasmaki, line 27: Is this `&.horizontal` empty for a reason?

frontend/src/components/CustomerWeightsVisualization.tsx
```diff
@@ -0,0 +1,152 @@
...
+            {percent(1).format(totalValue > 0 ? entry.value / totalValue : 0)}
+          </span>
+        </div>
+      ))}
+    </div>
+  );
+};
+
+const CustomerWeightsPercentageBar: FC<{
+  customerStakes: CustomerStakes[];
```
juhoeerola, line 61: `filter: opacity(70%)` style for the percentage bar in this component should work for the different look (the 70% is what was in figma)

juhoeerola (commented):
With the slightly transparent colour added for the weights, I think this is good

IidaKainu:
Aleksi is busy with the new project, so I rebased this on master and applied review fixes. Also noticed the PercentageBars were not appearing in roadmapGraph due to "css" missing in the vertical className.

juhoeerola (approved):
> PercentageBars were not appearing in roadmapGraph

Good catch

PatricKangasmaki (approved):

PR #382, opened 2022-04-20
Planner read-mode

IidaKainu:
Planner is accessible in read mode to users with Permission.VersionRead.
- Developer or business -users do not have Permission.RoadmapReadCustomerValues, so in those cases the customer weights page is hidden and unweighted milestone values is used
- Add VersionRead-permission to business role


juhoeerola (approved):
Looking good, and seems to work as intended 👍

PR #383, opened 2022-04-22
Change Checkbox-component label to optional

IidaKainu:
Small refactor to use the Checkbox-component in RegisterPage to avoid code duplication.


juhoeerola (approved):
🔥

PR #384, opened 2022-04-25
Add milestone selection to MilestonesEditor

IidaKainu:
CTRL + click supports multiselection.
Unify selection styles in RoadmapGraph and MilestonesEditor.

- Should the selection be cleared when milestones or tasks are dragged?

### DIFF
--- a/frontend/src/pages/MilestonesEditor.tsx
+++ b/frontend/src/pages/MilestonesEditor.tsx
@@ -1,4 +1,4 @@
-import { MouseEvent, useEffect, useState } from 'react';
+import { MouseEvent, KeyboardEvent, useEffect, useState } from 'react';
 import {
   DragDropContext,
   Draggable,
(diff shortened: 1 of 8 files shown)


frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -266,6 +270,20 @@ export const MilestonesEditor = () => {
     }
   };
 
+  const selectVersions = (e: MouseEvent | KeyboardEvent, id: number) => {
+    e.stopPropagation();
+    const containsClicked = selectedVersions.includes(id);
+    if (e.ctrlKey) {
```
juhoeerola, line 276:
```suggestion
    if (e.ctrlKey || e.metaKey) {
```


On mac ctrl-click triggers only the context menu, so this is impossible to reach with mouse.

On mac <kbd>ctrl</kbd> is usually replaced with <kbd>command</kbd> which sets the `e.metaKey` to true. That [*might* be](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey) the <kbd>⊞ Windows</kbd> on other platforms.

I think checking also the meta key would be better than special casing the context menu.

juhoeerola (approved):
Seems good, just needs another modifier key to work on macos

IidaKainu:
Added the modifier key

frontend/src/pages/MilestonesEditor.tsx
IidaKainu: Thanks for pointing this out. `⊞ Windows` works with the `e.metaKey`, but I do not see it as a problem as it is rarely used in web.

PR #388, opened 2022-04-27
Remove expiration from trello authorisation

juhoeerola:
https://developer.atlassian.com/cloud/trello/guides/rest-api/authorization/#1-authorize--route-options


IidaKainu (approved):
👍

PR #389, opened 2022-04-27
Fix AddTaskModal description field size

AnttiVaarala:


IidaKainu (approved):

frontend/src/components/modals/AddTaskModal.module.scss
```diff
@@ -0,0 +1,4 @@
+textarea {
+    min-height: 20em !important;
```
IidaKainu, line 2: Since the user can resize the area, I would add 20em as the default height with a possibility to go smaller. Not that users need a smaller field, it resizing just feels more fun (if that makes at all sense).

As a side note there the tab indent should be two spaces. Do you have automatic code formatting (with prettier) enabled on your editor? Let me know if you want setup help
```suggestion
  min-height: 5em !important;
  height: 20em;
```

IidaKainu (changes requested):
Just a small change suggestion

frontend/src/components/modals/AddTaskModal.module.scss
AnttiVaarala: It makes sense, I committed the changes.
I have a formatter (Prettier) but forgot to run it before committing.

frontend/src/components/modals/AddTaskModal.module.scss
```diff
@@ -0,0 +1,4 @@
+textarea {
+  min-height: 5em !important;
+  height: 20em;
+}
```
IidaKainu, line 4: Maybe still add the max-height back in, so a scrollbar won't appear in the modal on resize 👍 
```suggestion
  height: 20em;
  max-height: 30em;
}
```

IidaKainu (changes requested):

IidaKainu (approved):

PR #396, opened 2022-05-03
Change task rating scale

AnttiVaarala:


frontend/src/components/MetricsSummary.tsx
```diff
@@ -21,7 +21,7 @@ export const MetricsSummary: FC<MetricsProps> = ({
       <>
         <div className={classes(css.dash)} />
         <div className={classes(css.dataNumberWrapper)}>
-          {value}
+          {Math.round(Number(value))}
```
IidaKainu, line 24: Just to note: we'll need a null check once this is rebased, since in master the value can be null as well
```suggestion
          {value !== null && Math.round(Number(value))}
```

shared/utils/conversion.ts
```diff
@@ -0,0 +1,15 @@
+export function convertScale(rating: number): number;
+export function convertScale(rating: number | undefined): number | undefined;
+export function convertScale(rating: number | undefined) {
+  if (rating === undefined) return undefined;
+  else if (rating < 0) return rating;
+  return rating ** 2;
+}
```
IidaKainu, line 7: ConvertScale is not used with undefined values, so I'd change the definition
```suggestion
export const convertScale = (rating: number) =>
  rating < 0 ? rating : rating ** 2;
```

shared/utils/conversion.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function revertScale(rating: number): number;
+export function revertScale(rating: number | undefined): number | undefined;
+export function revertScale(rating: number | undefined) {
+  if (rating === undefined) return undefined;
+  else if (rating < 0) return rating;
+  return Math.sqrt(rating);
+}
```
IidaKainu, line 15: Same goes for revertScale
```suggestion
export const revertScale = (rating: number) =>
  rating < 0 ? rating : Math.sqrt(rating);
```

frontend/src/components/RatingBars.tsx
```diff
@@ -58,7 +67,7 @@ export const TaskRatingBar: FC<RatingBarProps> = ({
       {!readonly && (
         <div className={classes(css.rating)}>
           {hover > 0
-            ? hover
+            ? labels[hover]
```
IidaKainu, line 70: If `convertScale(hover)` is used there, labels are not needed

frontend/src/components/TaskHeatmap.tsx
```diff
@@ -78,13 +79,15 @@ export const TaskHeatmap = () => {
...
         </div>
-        <p className={classes(css.yaxismax)}>5</p>
+        <p className={classes(css.yaxismax)}>25</p>
         <p className={classes(css.yaxismin)}>1</p>
         <p className={classes(css.yaxislabel)}>Value</p>
       </div>
-      <p className={classes(css.xaxismin)}>1</p>
-      <p className={classes(css.xaxislabel)}>Complexity</p>
-      <p className={classes(css.xaxismax)}>5</p>
+      <div className={classes(css.xLabelContainer)}>
```
IidaKainu, line 86: Is this div needed? At least no styles for xLabelContainer exist

frontend/src/components/TaskHeatmap.tsx
```diff
@@ -78,13 +79,15 @@ export const TaskHeatmap = () => {
             </div>
           ))}
         </div>
-        <p className={classes(css.yaxismax)}>5</p>
+        <p className={classes(css.yaxismax)}>25</p>
```
IidaKainu, line 82: I'd probably use `{convertScale(5)}` for these numbers, just so it's easier to remember to change if needed

IidaKainu (changes requested):
A few suggestions, looks good otherwise

frontend/src/components/MetricsSummary.tsx
juhoeerola: Where was this problem, so that rounding was needed? I forgot and couldn't find it anymore

Not needing this would be great

frontend/src/components/TaskHeatmap.tsx
juhoeerola: Not sure if it would be better to also "convert" the ones for consistency, even though those stay the same

server/tests/taskratings.test.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from '../../shared/types/customTypes';
 import { withoutPermission } from './testUtils';
 import TaskRating from '../src/api/taskratings/taskratings.model';
+import { convertScale, revertScale } from '../../shared/utils/conversion';
```
juhoeerola, line 11: There are still quite a few raw, unconverted values. Each one should be converted first, even the `-1` would be good to convert

shared/utils/conversion.ts
juhoeerola: I'd keep the `undefined` check in, since the `$parseDatabaseJson` passes `any` to this.
And for some reason that is called few times with partial data that doesn't have the value.

juhoeerola (changes requested):

frontend/src/components/MetricsSummary.tsx
IidaKainu: I remember some metrics would get too large with the new scaling if decimals were shown too. Maybe it was in task overview. 
One thought is, instead of modifying this component, maybe we could do rounding in places where it's needed.

IidaKainu: Oh right the content can overlap in task map. What do you think about only rounding there and removing the statement from MetricsSummary?

IidaKainu (approved):
Good on my behalf

juhoeerola (approved):

frontend/src/components/MetricsSummary.tsx
```diff
@@ -4,9 +4,15 @@ import css from './MetricsSummary.module.scss';
 
 const classes = classNames.bind(css);
 
+const numFormat = (number: number) =>
+  new Intl.NumberFormat(undefined, {
+    maximumFractionDigits: 0,
+    ...(number < 10 && { maximumFractionDigits: 1 }),
```
juhoeerola, line 10:
```suggestion
    maximumFractionDigits: number < 10 ? 1 : 0,
```

juhoeerola (changes requested):

juhoeerola (approved):
Looking good 👍

PR #397, opened 2022-05-03
POC: Variable size list to fixed size

juhoeerola:
Changes the variable size list to a fixed size one.

Each task is given enough room for two lines of text. This is much snappier for many hundreds of tasks.
The magic numbers should be refactored.


IidaKainu (approved):
Looks good. However, consider removing the margins from .instructions (milestonesEditor) so the milestones look more similar on task drag

IidaKainu (approved):

PR #398, opened 2022-05-05
Add local saving of node positions on taskmap

PatricKangasmaki:
Saves the current node locations to localstorage. If localstorage is empty when opening task map, it will generate autolayout with Dagre.
Also added a reset button which clears the storage and applies dagre autolayout.
These functionalities required to update reactflow from v9 to v10


juhoeerola (changes requested):
Apparently the react-flow library added attribution (link in bottom right) in the new version. I guess this is fine, just pointing it out, and incase we want to position it somewhere else.

frontend/src/components/TaskMap.tsx
```diff
@@ -231,10 +311,12 @@ export const TaskMap: FC<{
...
         connectionLineComponent={ConnectionLine}
-        elements={flowElements}
-        nodeTypes={{
-          special: CustomNodeComponent,
-        }}
+        nodesDraggable={groupDraggable}
+        nodes={groups}
+        edges={edges}
+        onNodesChange={onNodesChange}
+        onNodeDragStop={onNodesDragStop}
```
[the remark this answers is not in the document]
PatricKangasmaki: I'll change the onNodesChange and onNodeDragStop 👍

> Same with nodeTypes and edgeTypes

React flow documentation suggests that the nodeTypes and edgeTypes should either be defined outside the component or alternatively they can be memoized so I followed that:

>Whenever you pass new node or edge types, we create wrapped node or edge component types in the background. This means that you should not create a new nodeType or edgeType object on every render. Memoize your nodeTypes and edgeTypes or define them outside of the component when they don't change

PatricKangasmaki:
Added all your suggestions, great points 👍 I marked this PR as draft until I can figure this out:

>Furthermore, I could replicate the bug about editing taskmap on another account and changes not being applied to the other taskmap
I managed to replicate this as well, I'll look into it

PatricKangasmaki:
>Furthermore, I could replicate the bug about editing taskmap on another account and changes not being applied to the other taskmap

Turns out this bug is linked to websocket connection issues (after logging in, the websocket will not update any changes done by other users/browsers before you refresh the page) and not taskmap directly so I'll reopen the PR.

IidaKainu (changes requested):
Small nitpicks, otherwise looks ready to go 👍

frontend/src/components/TaskMap.tsx
```diff
@@ -207,9 +225,18 @@ export const TaskMap: FC<{
...
     setIsLoading(false);
   };
 
+  const resetCanvas = () => {
+    localStorage.removeItem(flowKey);
+    drawCanvas();
+  };
+
+  const nodeTypes = useMemo(() => ({ special: CustomNodeComponent }), []);
+  const edgeTypes = useMemo(() => ({ custom: CustomEdge }), []);
```
juhoeerola, line 238: I guess it would make sense to define these outside the component as these are constant (no dependencies), and thus the `useMemo` isn't really needed for anything.

PatricKangasmaki: Good point 👍

PatricKangasmaki:
Added both suggestions 👍

IidaKainu (approved):
🔥

juhoeerola (approved):

PR #402, opened 2022-05-10
Task status edition

IidaKainu:
As a note, I gave up (at least for now) creating the generic/reusable withButtons due to time constraints for the UX test session. EditableSelect and -field were too different.

### DIFF
--- a/frontend/src/components/EditableSelect.tsx
+++ b/frontend/src/components/EditableSelect.tsx
@@ -1,0 +1,149 @@
+import classNames from 'classnames';
+import {
+  FC,
+  useState,
+  ComponentPropsWithoutRef,
+  Dispatch,
+  SetStateAction,
+  useEffect,
+} from 'react';
+import Select from 'react-select';
+import { CloseButton, ConfirmButton, EditButton } from './forms/SvgButton';
+import { LoadingSpinner } from './LoadingSpinner';
+import { FieldError } from './forms/FormField';
+import css from './EditableSelect.module.scss';
+
+const classes = classNames.bind(css);
+
+interface Option {
+  value: number;
+  label: string;
+}
+
+interface WithButtonsProps {
+  onOk: (newValue: number, fieldId: string) => Promise<string | void>;
+  value: Option;
+  fieldId: string;
+  format: string | undefined;
+  options: Option[];
+}
+
+const withButtons = (Component: typeof EditableSelect) => ({
+  onOk,
+  value,
+  fieldId,
+  format,
+  options,
+}: WithButtonsProps) => {
+  const [editOpen, setEditOpen] = useState(false);
+  const [editValue, setEditValue] = useState<Option>(value);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleConfirm = () => {
+    setIsLoading(true);
+    onOk(editValue.value, fieldId).then((returnedError) => {
+      if (returnedError) setErrorMessage(returnedError);
+      else setEditOpen(false);
+      setIsLoading(false);
+    });
+  };
+
+  const handleCancel = () => {
+    setErrorMessage('');
+    setEditValue(value);
+    setEditOpen(false);
+  };
+
+  return (
+    <div
+      className={classes(css.withButtonsContainer, {
+        [css.withButtonsContainerEditMode]: editOpen,
+      })}
+    >
+      <Component
+        id={fieldId}
+        editOpen={editOpen}
+        editValue={editValue}
+        onChange={(newValue) => {
+          setEditValue(newValue as Option);
+          setErrorMessage('');
+        }}
+        error={{ message: errorMessage, setMessage: setErrorMessage }}
+        isLoading={isLoading}
+        value={value}
+        format={format}
+        options={options}
+      />
+      {editOpen ? (
+        <div className={classes(css.buttonsDiv)}>
+          <CloseButton onClick={handleCancel} />
+          <ConfirmButton onClick={handleConfirm} />
+        </div>
+      ) : (
+        <div className={classes(css.editButtonDiv)}>
+          <EditButton
+            fontSize="medium"
+            onClick={() => {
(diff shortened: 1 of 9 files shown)


IidaKainu:
rebased on master

IidaKainu:
Disabled the search and the menu stays open after once the menu is opened. Could not get the menu to open automatically though. 
This is not a necessity for the UX test, so I can think of better options later.

IidaKainu:
> Could not get the menu to open automatically though.

Added a workaround for this.

frontend/src/pages/TaskOverviewPage.tsx
```diff
@@ -33,15 +33,33 @@ import { MissingRatings } from '../components/MissingRatings';
...
 const classes = classNames.bind(css);
 
 export const getTaskOverviewData = (
   task: Task,
   editable: boolean,
   users: RoadmapUser[],
+  roadmapId: number,
 ) => {
   const { value, complexity } = valueAndComplexitySummary(task);
+  const [patchTaskTrigger] = apiV2.usePatchTaskMutation();
```
juhoeerola, line 48: This is a hook, so it can't be used like this in a function. Furthermore, this is called conditionally, so there are some errors from different hooks at different renders.

frontend/src/components/EditableSelect.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+const withButtons = (Component: typeof EditableSelect) => ({
+  onOk,
+  value,
+  fieldId,
+  format,
+  options,
+}: WithButtonsProps) => {
+  const [editOpen, setEditOpen] = useState(false);
+  const [editValue, setEditValue] = useState<Option | undefined>(undefined);
```
juhoeerola, line 39: Would the edit value ever need to be `undefined`, or could it be initialised with `value`?

frontend/src/components/EditableSelect.tsx
```diff
@@ -0,0 +1,151 @@
...
+  isLoading,
+  format,
+  error,
+  ...props
+}) => {
+  const [menuIsOpen, setMenuIsOpen] = useState(false);
+
+  useEffect(() => {
+    if (editOpen) setMenuIsOpen(true);
+    else setMenuIsOpen(false);
```
juhoeerola, line 122:
```suggestion
    setMenuIsOpen(editOpen);
```

I don't quite understand why `menuIsOpen={true}` wouldn't work for the select, but at least this can be done bit more simply

frontend/src/components/EditableSelect.tsx
```diff
@@ -0,0 +1,151 @@
...
+          menuPortal: (base) => ({ ...base, zIndex: 9999 }),
+          indicatorsContainer: () => ({ paddingRight: '5% !important' }),
+        }}
+        value={editValue}
+        menuPortalTarget={document.body}
+        options={options}
+        isSearchable={false}
+        menuIsOpen={menuIsOpen}
+      />
+      <FieldError msg={error.message} />
```
juhoeerola, line 146: I'm not quite sure about this error display.

It's rendered to the side of the select, in a quite narrow space.
Also, the only error should happen when confirming the choice, but the select closes without showing any error.

frontend/src/components/EditableSelect.tsx
```diff
@@ -0,0 +1,151 @@
...
+        error={{ message: errorMessage, setMessage: setErrorMessage }}
+        isLoading={isLoading}
+        value={value}
+        format={format}
+        options={options}
+      />
+      {editOpen ? (
+        <div className={classes(css.buttonsDiv)}>
+          <CloseButton onClick={handleCancel} />
+          {errorMessage === '' && <ConfirmButton onClick={handleConfirm} />}
```
juhoeerola, line 81:
```suggestion
          <ConfirmButton onClick={handleConfirm} />
```
This isn't doing any form validation (each option is valid), so only errors should be from submitting the choice.
If any validation, it could be checked here that `editValue !== undefined`.

frontend/src/pages/TaskOverviewPage.tsx
```diff
@@ -190,9 +228,8 @@ const TaskOverview: FC<{
         name={task.name}
         previousAndNext={siblingTasks}
         onOverviewChange={(id) => history.push(`${tasksPage}/${id}`)}
-        onDataEditConfirm={handleEditConfirm}
         key={task.id}
-        {...getTaskOverviewData(task, hasEditPermission, users!)}
+        {...getTaskOverviewData(task, hasEditPermission, users!, roadmapId)}
```
juhoeerola, line 232: The `users` could also be checked at the call site, or alternatively the function could take optional users (which might be nicer as the users aren't so important for it).

juhoeerola (changes requested):
Noticed an issue with hook use, and the error display doesn't seem useful at the moment.

Other than that, just couple small suggestions.
And it *would* be nice to share some of the common code between this and the `EditableText`, but I think copy-paste is fine for now.

frontend/src/components/EditableSelect.tsx
IidaKainu: Good catch. I did not notice the styles.

At some point the error (in EditableText) showed request errors too, which was my intended use case. However, the page is reloaded whether the request fails or not. I suppose apiV2 patchTask invalidates Tasks-tag and reload happens due to tasks being fetched. It would be nice to invalidate tags only on succeeding requests, but I don't think it's possible... so currently errors is never shown in editableSelect. Not sure about what to do 🤔

juhoeerola (changes requested):
System for showing errors that is never actually shown to the user is pretty useless, I think we should think something else for it or just remove it. (unless conditional tag invalidation works)

Other than that, the feature seems to be working fine
...so, more nitpicky notes this time

frontend/src/components/EditableSelect.tsx
IidaKainu: This works, great!

IidaKainu:
Added all fixes, thank you for the detailed review 👍

PR #405, opened 2022-05-11
Fix planner crash when adding a milestone

juhoeerola:


frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -473,7 +474,7 @@ export const MilestonesEditor = () => {
                         {completed && <DoneAll />}
                         {version.name}
                       </div>
-                      {versionLists[version.id]?.length === 0 ? (
```
juhoeerola, line 476: This was the issue, comparing `undefined === 0` when the milestone was just added.

frontend/src/pages/MilestonesEditor.tsx
```diff
@@ -515,15 +516,13 @@ export const MilestonesEditor = () => {
                       </div>
                       <div className={classes(css.summaryWrapper)}>
                         <MilestoneRatingsSummary
-                          tasks={versionLists[version.id] || []}
+                          tasks={list}
                           completed={completed}
                         />
                       </div>
-                      {!completed && !!versionLists[version.id].length && (
```
IidaKainu, line 522: I'd still leave the length check in, since we decided not to show MilestoneCompletedness on empty milestones.
image

IidaKainu (changes requested):
Crashing is fixed, but functionality has been changed

frontend/src/pages/MilestonesEditor.tsx
juhoeerola: Yeah, missed this at first
Added the length check back in

IidaKainu (approved):
👍

PR #406, opened 2022-05-13
Remove task map task ratings

IidaKainu:
https://trello.com/c/ojE7isyv/660-taskien-arvot-pois-task-map-listauksesta
Also all tasks are changed to same height, removing the need for calculating task name heights.


PatricKangasmaki (approved):
Looking good 👍

PR #409, opened 2022-05-13
Planner relation modal in drawer

juhoeerola:
Showing the relations in a drawer in the side instead of modal that would block the rest of the page.

I'm maybe abusing the modal link system a bit, but it works.
There is an extra horizontal scroll bar when the drawer is closed for some reason, don't yet know why.

### DIFF
--- a/frontend/src/components/TaskRelationTable.tsx
+++ b/frontend/src/components/TaskRelationTable.tsx
@@ -1,8 +1,7 @@
-import { FC, useRef, useState, useEffect } from 'react';
+import { FC, useState, useEffect, useCallback } from 'react';
 import { useSelector } from 'react-redux';
 import { skipToken } from '@reduxjs/toolkit/query/react';
 import { Trans } from 'react-i18next';
-import { VariableSizeList } from 'react-window';
 import Select from 'react-select';
 import classNames from 'classnames';
 import CachedIcon from '@mui/icons-material/Cached';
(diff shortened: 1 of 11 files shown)


frontend/src/components/modals/RelationsModal.tsx
```diff
@@ -126,37 +126,27 @@ const relationTable: (
...
+            showMilestoneName: showMilestoneNames,
+          }}
+          Action={({ task: { id } }) => (
+            <div className={classes(css.relationAlert)}>
+              <BadRelationWarning
+                relation={{ to: id, type }}
+                badRelations={badRelations}
+              />
+            </div>
+          )}
```
IidaKainu, line 148: This is highlighted with a react/no-unstable-nested-components -warning. The component could be memoized

`const BadRelation = useCallback(
      ({ task: { id } }) => (
        <div className={classes(css.relationAlert)}>
          <BadRelationWarning
            relation={{ to: id, type }}
            badRelations={badRelations}
          />
        </div>
      ),
      [badRelations],
    );`

frontend/src/components/TaskRelationTable.tsx
```diff
@@ -129,41 +109,25 @@ const relationTable: (def: RelationTableDef) => FC<RelationTableProps> = ({
...
+              e.stopPropagation();
+              if (roadmapId) {
+                removeTaskRelation({
+                  roadmapId,
+                  relation: buildRelation(task.id, id),
+                });
+              }
+            }}
+          />
+        )}
```
IidaKainu, line 129: The component here could be memoized with useCallback as well

IidaKainu (changes requested):
The TaskTable looks great 👍 

> I'm maybe abusing the modal link system a bit, but it works.

Maybe. However, the system seems to work fine for multiple purposes, so it could be renamed to something more general(?)

More specifically, the relationsModal is built like a modal, but is used as a drawer. This leads to the modal footer buttons looking out of place in a drawer. If you'd like to keep the RelationsModal flexible, maybe show the footer depending on the ModalAction type. Otherwise, the ModalFooter could simply be removed.

IidaKainu (approved):

PR #410, opened 2022-05-13
Fix crash on missing tasks in milestone editor

juhoeerola:
The `tasks` could be `undefined` in some circumstances, which lead to
silent failure in lookup from `tasksById`, and a crash later in the
page.

Repro of the error:
- navigate to `/roadmap/1/planner/graph`
- refresh the page
- select the `milestones` tab
- page crashes


PatricKangasmaki (approved):
👍👍

PR #411, opened 2022-05-14
Gitlab integration

JimiRantanen:
~~A work in progress, things to add:~~
- ~~Access token refresh using refresh token (access tokens expire in 2 hours)~~
- ~~Task filters when importing, like in other integrations~~

~~Also, I'd like to refactor out the million toString() calls I had to use - GitLab's id values are numbers and they didn't convert to strings easily~~

### DIFF
--- a/frontend/src/components/IntegrationConfiguration.tsx
+++ b/frontend/src/components/IntegrationConfiguration.tsx
@@ -176,13 +176,15 @@ const MapStates: FC<{ configuration: IntegrationConfiguration }> = ({
 };
 
 const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
-  configuration: { id, name, roadmapId, boardId },
+  configuration,
 }) => {
+  const { id, name, roadmapId, boardId } = configuration;
   const { t } = useTranslation();
 
   const [errorMessage, setErrorMessage] = useState('');
   const [patch] = apiV2.usePatchIntegrationConfigurationMutation();
 
+  const [boardsTokenError, setBoardsTokenError] = useState(false);
   const [tokenError, setTokenError] = useState(false);
   const selected = apiV2.useGetIntegrationSelectedBoardQuery(
     { name, roadmapId },
@@ -197,15 +199,23 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
   const isTokenError = (err: any) => err?.data?.error === 'InvalidTokenError';
 
   useEffect(() => {
-    setTokenError(selected.isSuccess && isTokenError(boards.error));
-  }, [boards, selected]);
+    setBoardsTokenError(selected.isSuccess && isTokenError(boards.error));
+    if (
+      selected.isError &&
+      (isTokenError(selected.error) || isTokenError(boards.error))
+    ) {
+      setTokenError(true);
+      setErrorMessage(t('Token is expired'));
+    }
+  }, [boards, selected, t]);
 
-  if (selected.isError) return null;
   if (
     !selected.isLoading &&
     !selected.data &&
+    !selected.error &&
     !boards.isLoading &&
-    !boards.data
+    !boards.data &&
+    !boards.error
   )
     return null;
 
@@ -218,11 +228,16 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
         <LoadingSpinner />
       ) : (
         <>
-          {(tokenError || errorMessage) && (
+          {(tokenError || boardsTokenError || errorMessage) && (
             <Alert
               severity={errorMessage ? 'error' : 'info'}
               onClose={() => {
-                if (tokenError) setTokenError(false);
+                if (boardsTokenError) setBoardsTokenError(false);
+                if (tokenError) {
+                  boards.refetch();
+                  selected.refetch();
+                  setTokenError(false);
+                }
                 if (errorMessage) setErrorMessage('');
               }}
               icon={false}
@@ -230,37 +245,42 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
               {errorMessage || t('Oauth can not select board')}
             </Alert>
           )}
-          <Select
-            name="board"
-            id="board"
-            className="react-select"
-            classNamePrefix="react-select"
-            styles={{ menuPortal: (base) => ({ ...base, zIndex: 9999 }) }}
-            placeholder="Select board"
-            isDisabled={boards.isError}
-            menuPortalTarget={document.body}
-            onChange={(entry) => {
-              if (entry) {
-                patch({ id, name, roadmapId, boardId: entry.value })
-                  .unwrap()
-                  .catch((err) => {
-                    setErrorMessage(
-                      err.data?.message ?? 'something went wrong',
-                    );
-                  });
+          {!boards.isError && !selected.isError && (
+            <Select
+              name="board"
+              id="board"
+              className="react-select"
+              classNamePrefix="react-select"
+              styles={{ menuPortal: (base) => ({ ...base, zIndex: 9999 }) }}
+              placeholder="Select board"
+              isDisabled={boards.isError}
+              menuPortalTarget={document.body}
+              onChange={(entry) => {
+                if (entry) {
+                  patch({ id, name, roadmapId, boardId: entry.value })
+                    .unwrap()
+                    .catch((err) => {
+                      setErrorMessage(
+                        err.data?.message ?? 'something went wrong',
+                      );
+                    });
+                }
+              }}
+              defaultValue={
+                selected.data && {
+                  value: selected.data.id,
+                  label: selected.data.name,
+                }
               }
-            }}
-            defaultValue={
-              selected.data && {
-                value: selected.data.id,
-                label: selected.data.name,
-              }
-            }
-            options={boards.data?.map((board) => ({
-              value: board.id,
-              label: board.name,
-            }))}
-          />
+              options={boards.data?.map((board) => ({
+                value: board.id,
+                label: board.name,
+              }))}
+            />
+          )}
+          {boardId && !boards.isError && !selected.isError && (
+            <MapStates configuration={configuration} />
+          )}
         </>
       )}
     </div>
@@ -468,9 +488,6 @@ export const IntegrationConfig: FC<{ roadmapId: number; name: string }> = ({
               <Oauth configuration={configuration} />
               <SelectBoard configuration={configuration} />
             </>
-          )}
-          {configuration?.boardId && (
-            <MapStates configuration={configuration} />
           )}
           {configuration && <Remove configuration={configuration} />}
         </div>
--- a/frontend/src/components/IntegrationConfiguration.tsx
+++ b/frontend/src/components/IntegrationConfiguration.tsx
@@ -200,10 +200,7 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
 
   useEffect(() => {
     setBoardsTokenError(selected.isSuccess && isTokenError(boards.error));
-    if (
-      selected.isError &&
-      (isTokenError(selected.error) || isTokenError(boards.error))
-    ) {
+    if (selected.isError && isTokenError(selected.error)) {
       setTokenError(true);
       setErrorMessage(t('Token is expired'));
     }
@@ -245,7 +242,7 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
               {errorMessage || t('Oauth can not select board')}
             </Alert>
           )}
-          {!boards.isError && !selected.isError && (
+          {!selected.isError && (
             <Select
               name="board"
               id="board"
--- a/frontend/src/components/IntegrationConfiguration.tsx
+++ b/frontend/src/components/IntegrationConfiguration.tsx
@@ -193,7 +193,7 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
 
   const boards = apiV2.useGetIntegrationBoardsQuery(
     { name, roadmapId },
-    { skip: !selected.isSuccess },
+    { skip: !!boardId && !selected.isSuccess },
   );
 
   const isTokenError = (err: any) => err?.data?.error === 'InvalidTokenError';
@@ -241,7 +241,7 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
               {errorMessage || t('Oauth can not select board')}
             </Alert>
           )}
-          {selected.isSuccess && (
+          {!selected.isError && (
             <Select
               name="board"
               id="board"
@@ -483,9 +483,9 @@ export const IntegrationConfig: FC<{ roadmapId: number; name: string }> = ({
             <>
               <Oauth configuration={configuration} />
               <SelectBoard configuration={configuration} />
+              <Remove configuration={configuration} />
             </>
           )}
-          {configuration && <Remove configuration={configuration} />}
         </div>
       )}
     </div>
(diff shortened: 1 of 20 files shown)


frontend/src/pages/OauthRedirectPage.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <Redirect
+      to={`/roadmap/${query.get(
+        'roadmapId',
+      )}/configure?openModal=SETUP_OAUTH_MODAL&modalProps=%7B%22name%22%3A%22${query.get(
+        'integrationName',
+      )}%22%2C%22roadmapId%22%3A${query.get(
+        'roadmapId',
+      )}%2C%22code%22%3A%22${query.get('code')}%22%7D`}
+    />
```
juhoeerola, line 21:
```suggestion
  const roadmapId = Number(query.get('roadmapId')!);
  return (
    <Redirect
      to={`/roadmap/${roadmapId}/configure${modalLink(ModalTypes.SETUP_OAUTH_MODAL, {
        roadmapId,
        name: query.get('integrationName')!,
        code: query.get('code')!
      }}
    />
```

Could use the `modalLink` function to build the query params. It's more readable and also checks the types.

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,169 @@
...
+  getLastUsedRedirectUri() {
+    return this.lastUsedRedirectUri;
+  }
+
+  async boards() {
+    const response = await this.fetch<{ id: string; name: string }[]>('boards');
+    return (response as any[]).map((board) => ({
+      id: board.id as string,
+      name: board.name as string,
+    }));
```
juhoeerola, line 99:
```suggestion
    const response = await this.fetch<{ id: number; name: string }[]>('boards');
    return response.map((board) => ({
      id: `${board.id}`,
      name: board.name,
    }));
```

I suppose this is the root cause for the `toString()`s. Here practically `any` is coerced to `{ id: string; name: string }[]`, and I suppose the `id` isn't actually a `string`.

So the correct place to convert would be the boundary from `any` to a known type.


Same might happen with `columns()` and others, if there are more type mismatches.

server/src/api/integration/integration.controller.ts
```diff
@@ -78,6 +78,56 @@ export const postConfigurations: RouteHandlerFnc = async (ctx) => {
...
+          }
+        };
+        await upsertToken('access_token', accessToken);
+        await upsertToken('refresh_token', refreshToken);
+      });
+      return await resourceCall();
+    } else throw err;
+  }
+};
+
```
JimiRantanen, line 130: When the token expires, this approach ends with one call returning InvalidTokenError (token expired) and the other two returning InvalidGrantError (the token is revoked as a new one has been fetched with refresh).  After a single set of fails, the token does get successfully refreshed and subsequent calls work, but it is a bit clunky right now.

Any ideas @juhoeerola, a refresh button on the config page or something else?

juhoeerola: I think we could sequence the api calls requiring oauth, such that one call must succeed before the others are even tried. This should handle the normal use cases.

There will still be cases where the token could expire while the user is on the page, or when multiple users trigger the refresh at the same time. However, I think regular page refresh (instead of explicit "refresh token" button) could be enough for the cases when the automatic refresh fails. (A token error could show error message suggesting a page refresh.)

JimiRantanen:
How about this? 2 requests are still done at the same time (boards and the selected board), but it shouldn't be an issue - If they both fail and at least one of them is an InvalidTokenError, an alert is shown stating the token is invalid, and closing it tries refetching the boards and the selected board. If the backend encounters an InvalidTokenError when fetching, it silently refreshes the token if the integration has a token refresh method available.

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,213 @@
...
+        );
+        resolve({
+          url: `https://${config.host}${authorizePath}?client_id=${config.consumerkey}&response_type=code&state=${state}&scope=read_api&redirect_uri=${redirectUri}`,
+          token: '',
+          tokenSecret: '',
+        });
+      }),
+    swapOAuthToken: ({ verifierToken, token, tokenSecret, roadmapId }) =>
+      new Promise((resolve, reject) => {
+        const redirectUri = `${redirectBaseURL}?roadmapId=${roadmapId}&integrationName=gitLab`;
```
juhoeerola, line 47: This could make sense as a (top level) function
```ts
const redirectUri = (roadmapId: number) =>
  `${redirectBaseURL}?roadmapId=${roadmapId}&integrationName=gitLab`;
```

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,213 @@
...
+                reject(new InvalidTokenError(err.data));
+              else if (err.statusCode === 400 && /grant/i.test(err.data))
+                reject(new InvalidGrantError(err.data));
+              else reject(err);
+            } else
+              resolve({
+                accessToken: access_token,
+                refreshToken: refresh_token,
+              });
+          },
```
juhoeerola, line 71:
```suggestion
          (err, accessToken, refreshToken) => {
            if (err) reject(convertError(err));
            else resolve({ accessToken, refreshToken });
          },
```

Couple suggestions:
- the error check would be good to extract to a function (something like below)
- the parameters can be freely named, so could use the field names directly, as those also are camelCase which is the style convention
- the unused (trailing) parameter should be removed

```ts
const convertError = (err: { statusCode: number; data?: any }) => {
  if (err.statusCode === 401 && /token/i.test(err.data))
    return new InvalidTokenError(err.data);
  if (err.statusCode === 400 && /grant/i.test(err.data))
    return new InvalidGrantError(err.data);
  return err;
};
```

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,213 @@
...
+        );
+      }),
+  }),
+};
+
+class GitLabImporter implements IntegrationProvider {
+  private readonly projectId: string;
+  private readonly oauth2: OAuth2;
+  private readonly accessToken: string;
+  private lastUsedRedirectUri: string;
```
juhoeerola, line 110: This seems unused
```suggestion
```

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,213 @@
...
+  setLastUsedRedirectUri(uri: string) {
+    this.lastUsedRedirectUri = uri;
+  }
+  getLastUsedRedirectUri() {
+    return this.lastUsedRedirectUri;
+  }
+
+  async boards() {
+    const response = await this.fetch<{ id: number; name: string }[]>('boards');
+    return (response as any[]).map((board) => ({
```
juhoeerola, line 131:
```suggestion
    return response.map((board) => ({
```

The type cast is unnecessary here

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,213 @@
...
+      name: board.name,
+    }));
+  }
+
+  // In GitLab, columns ("lists") can only be created out of labels.
+  // The columns themselves do have an id, but they are named after the label.
+  // Therefore, this returns the column id and the label's name.
+  async columns(boardId: string) {
+    const lists = await this.fetch<
+      Array<{ id: number; label: { name: string } }>
```
juhoeerola, line 142:
```suggestion
     { id: number; label: { name: string } }[]
```
Could use consistently the same notation for array, unless there is some specific reason

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,213 @@
...
+    >(`boards/${boardId}/lists`);
+    return lists.map((list) => ({ id: `${list.id}`, name: list.label.name }));
+  }
+
+  async labels(boardId: string) {
+    const lists = await this.fetch<Array<{ label: { name: string } }>>(
+      `boards/${boardId}/lists`,
+    );
+    return lists.map((list) => list.label.name);
```
juhoeerola, line 151:
```suggestion
    const lists = await this.columns(boardId);
    return lists.map((list) => list.name);
```

Could use the `columns()` method to make the connection clear

server/src/api/gitLab/gitLabIntegration.ts
```diff
@@ -0,0 +1,213 @@
...
+      this.oauth2.get(
+        `https://gitlab.com/api/v4/projects/${this.projectId}/${resource}`,
+        this.accessToken,
+        (err, result) => {
+          if (!err) return resolve(JSON.parse(result as string));
+          if (err.statusCode === 401 && /token/i.test(err.data))
+            reject(new InvalidTokenError(err.data));
+          else if (err.statusCode === 400 && /grant/i.test(err.data))
+            reject(new InvalidGrantError(err.data));
+          else reject(err);
```
juhoeerola, line 195:
```suggestion
          if (err) reject(convertError(err));
          else resolve(JSON.parse(result as string));
```

Could keep it consistent

This is more obvious when the error check is refactored

juhoeerola (changes requested):
I didn't yet check the frontend

frontend/src/components/IntegrationConfiguration.tsx
```diff
@@ -218,49 +228,59 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
...
-                      err.data?.message ?? 'something went wrong',
-                    );
-                  });
-              }
-            }}
-            defaultValue={
-              selected.data && {
-                value: selected.data.id,
-                label: selected.data.name,
+          {!selected.isError && (
```
juhoeerola, line 245:
```suggestion
          {selected.isSuccess && (
```

I suppose `isSuccess` could be a better condition, as it can't really be error before the first load, but it doesn't have any data then either.

Same could work better for the skip condition, as that should sequence the call only after other successful call

```ts
  const boards = apiV2.useGetIntegrationBoardsQuery(
    { name, roadmapId },
    { skip: !selected.isSuccess },
  );
```

frontend/src/components/IntegrationConfiguration.tsx
```diff
@@ -218,49 +225,59 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
...
-              label: board.name,
-            }))}
-          />
+              options={boards.data?.map((board) => ({
+                value: board.id,
+                label: board.name,
+              }))}
+            />
+          )}
+          {boardId && !boards.isError && !selected.isError && (
```
juhoeerola, line 278:
```suggestion
          {boardId && selected.isSuccess && (
```

This shouldn't require all boards, only that a board is selected. The selected board query could work as a "token refresher" before rendering this, if the refresh would be added to that on the server

juhoeerola (changes requested):
Noticed one more bug in the backend: the token handling in the refresh and in the `integrationProvider` calls.

Other than that, the backend seems good. I'm not sure if it's good to leave out the refresh from get selected board, but if it works better without that's fine.

Also checked the frontend now, it seemed mostly solid, just the one incorrect precondition and the possible improvement from `isError` -> `!isSuccess`

JimiRantanen:
The problem with token refresh in both boards and the selected board calls is, that it can't be predicted which resolves first. I took the refresh out so it's always the selected board call that fails in case of an expired token, which also differentiates it from the another-user-oauth error, where it's always the boards call that fails. (They both fail with InvalidTokenError.)

This approach has a problem, where the "try to refresh by closing this error message" error will be displayed if another user goes to the settings page and tries using the integration, and as the boards call fails before it gets to the refresh part and the selected call doesn't refresh, the token never gets refreshed in this case. I'm trying to think of a better solution...

JimiRantanen:
Based on my testing, the skip option didn't work like that - both calls always fired.

juhoeerola:
> Based on my testing, the skip option didn't work like that - both calls always fired.

Odd, on my testing `skip: !selected.isSuccess` worked fine, firing the boards query only after the selected had returned from the backend.

What comes to "really specific" timings, those will happen, but I don't think we should worry about it too much as long as the user has some way to try again.

JimiRantanen:
I'll try it again - actually, after checking the file, it seems to check for `selected.isError` (instead of `!selected.isSuccess`) currently, which obviously wouldn't work. I'll try it with the proper check.

JimiRantanen:
After looking through the rest of the comments, I noticed you pointed out the very same thing.

Now it seems to work quite well, the skip parameter was very useful! I added token refresh to the selected call as well now. Now, the frontend tries to first fetch the selected board, and only if that call succeeds will the frontend try to fetch all boards. If the first attempt fails, the backend refreshes the token silently and tries again - the frontend sees this as a single successful request, no user interaction needed. After a successful selected board response, boards are fetched, and at this point the token has been refreshed (as the ok response of the refreshing call is needed before the boards call is even tried).

frontend/src/components/IntegrationConfiguration.tsx
```diff
@@ -176,36 +176,43 @@ const MapStates: FC<{ configuration: IntegrationConfiguration }> = ({
...
+  const [selectedError, setSelectedError] = useState(false);
   const selected = apiV2.useGetIntegrationSelectedBoardQuery(
     { name, roadmapId },
     { skip: !boardId },
   );
 
   const boards = apiV2.useGetIntegrationBoardsQuery(
     { name, roadmapId },
-    { skip: selected.isError },
+    { skip: !selected.isSuccess },
```
juhoeerola, line 196:
```suggestion
    { skip: !!boardId && !selected.isSuccess },
```

Noticed that this fails if when there is no board selected yet.

If there is no board selected, the selected board query is skipped, but this should be tried.

The queries should still remain sequential, since the selected query was skipped

frontend/src/components/IntegrationConfiguration.tsx
```diff
@@ -218,49 +225,58 @@ const SelectBoard: FC<{ configuration: IntegrationConfiguration }> = ({
...
-                      err.data?.message ?? 'something went wrong',
-                    );
-                  });
-              }
-            }}
-            defaultValue={
-              selected.data && {
-                value: selected.data.id,
-                label: selected.data.name,
+          {selected.isSuccess && (
```
juhoeerola, line 244:
```suggestion
          {!selected.isError && (
```

Sorry, this was my mistake in the previous review. I then realised that `selected.isSuccess` is too strict.
I guess `!selected.isError` may not be the most specific precondition for displaying this, but it should be reasonably good.

frontend/src/components/IntegrationConfiguration.tsx
```diff
@@ -468,9 +485,6 @@ export const IntegrationConfig: FC<{ roadmapId: number; name: string }> = ({
               <SelectBoard configuration={configuration} />
             </>
```
juhoeerola, line 486:
```suggestion
              <SelectBoard configuration={configuration} />
              <Remove configuration={configuration} />
            </>
```

The remove button could be moved in to the same group, as it has the same precondition.

juhoeerola (changes requested):
One more bug partly from my review yesterday 🤦, but with that fixed I think this should be ready and working.

server/src/api/integration/integration.controller.ts
```diff
@@ -70,16 +72,81 @@ export const postConfigurations: RouteHandlerFnc = async (ctx) => {
...
+const withTokenRefresh = async <T>(
+  ctx: Parameters<RouteHandlerFnc>[0],
+  resourceCall: () => Promise<T>,
+) => {
+  try {
+    return await resourceCall();
+  } catch (err) {
+    const provider = await oAuthProvider(ctx);
+    if (!(err instanceof InvalidTokenError && provider.tokenRefresh)) throw err;
```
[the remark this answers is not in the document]
juhoeerola: I think we can add this later, if we find out it would be needed

juhoeerola (approved):
I added fix for the last bug.

Check that I didn't botch anything more, and then this should be good to go

PR #412, opened 2022-05-16
Small rating table fixes

IidaKainu:
Render newlines and set max height higher

### DIFF
--- a/frontend/src/components/RatingTable.module.scss
+++ b/frontend/src/components/RatingTable.module.scss
@@ -72,14 +72,23 @@
 }
 
 .comment {
-  background-color: white;
+  background-color: $COLOR_WHITE;
   color: $COLOR_BLACK60;
+
   padding: 15px 20px;
   margin-top: 15px;
   border-radius: 10px;
+
   text-align: start;
+  white-space: pre-line;
+  max-height: 250px;
+  overflow: auto;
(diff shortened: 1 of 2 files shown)


juhoeerola:
I noticed that there seems to be no limit for the height of the comment box.
Now that newlines are rendered the box is more likely to grow "too much" vertically, so maybe we could limit it a bit.

IidaKainu:
Added a limit, hopefully it is not too restricting.

frontend/src/components/RatingTable.module.scss
```diff
@@ -72,14 +72,23 @@
...
   color: $COLOR_BLACK60;
+
   padding: 15px 20px;
   margin-top: 15px;
   border-radius: 10px;
+
   text-align: start;
+  white-space: pre-line;
+  max-height: 250px;
+  overflow: hidden;
```
juhoeerola, line 85: > hopefully it is not too restricting

Why not make it scrollable?

```suggestion
  overflow-y: auto;
```

juhoeerola (changes requested):

frontend/src/components/RatingTable.module.scss
IidaKainu: Good point

juhoeerola (approved):
👍

PR #413, opened 2022-05-16
Add sorting to rating table

IidaKainu:
https://trello.com/c/6f3SfpSE/661-taskoverview-ratings-listaan-sorttaus-asiakkaan-mukaan

User can sort ratings by customer, creator or value, with customer being the default.


juhoeerola (changes requested):

IidaKainu:
I am a bit embarrassed how messily this was done. Now the table is actually sorted by customer as the default. TaskRatingSortingTypesToText is fixed too

frontend/src/components/RatingTable.tsx
```diff
@@ -66,16 +74,24 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
...
   const typeString =
     type === TaskRatingDimension.BusinessValue ? 'value' : 'complexity';
   const roadmapId = useSelector(chosenRoadmapIdSelector);
 
+  const sorted = useMemo(() => sort(ratings), [ratings, sort]);
+
+  useEffect(() => {
+    if (!sorting.type.get())
+      sorting.type.set(TaskRatingSortingTypes.SORT_FOR_CUSTOMER);
+  }, [sorting.type]);
```
juhoeerola, line 90: `if (!sorting.type.get())` is bit risky, as the value can be `0`, not just `undefined`. In this case `0` happens to also be `SORT_FOR_CUSTOMER` but this is waiting for problems.

I'd rather add a default sort ordering parameter for `useSorting`. Something like
```ts
export const useSorting = <SortingType, ItemType, Default extends SortingType>(
  getSort: (t: SortingType | undefined) => SortBy<ItemType>,
  defaultType?: Default,
) => {
  const [type, setType] = useState<SortingType | undefined>(defaultType);
```

This can also be useful in other places, and gets rid of the `useEffect`

frontend/src/components/RatingTable.tsx
```diff
@@ -117,10 +133,39 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
...
+              )
+              .map((value) => (
+                <div
+                  className={classes(css.dropItem)}
+                  key={value}
+                  onClick={() => onSortingChange(value)}
+                  onKeyPress={() => onSortingChange(value)}
+                  role="button"
+                  tabIndex={0}
+                >
```
juhoeerola, line 163: Could change the `div` to an actual `button`.

This has at least couple advantages:
- `tabIndex` for free
- only needs to specify the selection action once (thus, I'd inline it)
- keyboard access
  - for free
  - more sane (only on <kbd>Space</kbd> and <kbd>Enter</kbd>, and closes on selection)

```suggestion
                <button
                  className={classes(css.dropItem)}
                  key={value}
                  onClick={() => sorting.type.set(value)}
                  type="button"
                >
```

Adding the following to `.dropItem` class seems to get the styles correct for buttons also:
```scss
width: 100%;
background-color: $COLOR_WHITE;
```

juhoeerola (changes requested):
Could also add translations

IidaKainu:
Added fixes

frontend/src/components/RatingTable.tsx
```diff
@@ -121,6 +135,30 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
     <div className={classes(css.ratingContainer)}>
       <div className={classes(css.titleContainer)}>
         <h3>{titleCase(t(typeString))}</h3>
+        <div className={classes(css.rightSide)}>
+          <Trans i18nKey="List order" />
+          <Dropdown
+            css={css}
+            title={TaskRatingSortingTypesToText(sorting.type.get())}
```
juhoeerola, line 142:
```suggestion
            title={t(TaskRatingSortingTypesToText(sorting.type.get()))}
```

frontend/src/components/RatingTable.tsx
```diff
@@ -121,6 +135,30 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
...
+                  typeof sortingType !== 'string',
+              )
+              .map((value) => (
+                <button
+                  className={classes(css.dropItem)}
+                  key={value}
+                  onClick={() => sorting.type.set(value)}
+                  type="button"
+                >
+                  {TaskRatingSortingTypesToText(value)}
```
juhoeerola, line 157:
```suggestion
                  {t(TaskRatingSortingTypesToText(value))}
```

juhoeerola (approved):

PR #417, opened 2022-05-17
Task map - remove remaining ratings

juhoeerola:
Not sure if this was intentionally left out from #406, but I now removed the ratings from the unstaged list too.

Also removed the search threshold from the task list, as that is fast enough.


IidaKainu (approved):
> Not sure if this was intentionally left out from https://github.com/Vincit/VISDOM-Roadmapper/pull/406

I probably forgot: this is better 👍

PR #420, opened 2022-05-18
Increase the size of body mini -font

IidaKainu:


juhoeerola (approved):
Not sure what would be the ideal font size, but this seems fine at least for my eyes

PR #421, opened 2022-05-18
Add edition for roadmap name and description

IidaKainu:
Takes advantage of overviewContent. Not necessarily in align with the look of the settings-page, but gets the job done.

### DIFF
--- a/frontend/src/pages/ConfigurationPage.tsx
+++ b/frontend/src/pages/ConfigurationPage.tsx
@@ -1,4 +1,4 @@
-import { Trans } from 'react-i18next';
+import { Trans, useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
 import { skipToken } from '@reduxjs/toolkit/query/react';
 import classNames from 'classnames';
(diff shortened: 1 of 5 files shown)


PatricKangasmaki (approved):
😄

frontend/src/pages/ConfigurationPage.tsx
```diff
@@ -22,13 +24,27 @@ const RoadmapConfigurationPageComponent = ({
...
   const roadmapId = useSelector(chosenRoadmapIdSelector);
   const dispatch = useDispatch<StoreDispatchType>();
   const userType = getType(userInfo, roadmapId);
-
+  const [patchRoadmap] = apiV2.usePatchRoadmapMutation();
   const { data: integrations } = apiV2.useGetIntegrationsQuery(
     roadmapId ?? skipToken,
   );
+  const { data: roadmaps } = apiV2.useGetRoadmapsQuery();
+  const roadmap = roadmaps?.find(({ id }) => id === roadmapId);
```
juhoeerola, line 36:
```suggestion
  const { data: roadmap } = apiV2.useGetRoadmapsQuery(undefined, {
    skip: !roadmapId,
    ...selectById(roadmapId),
  });
```

Could select by id directly from the query, and skip when no id is present.

juhoeerola (changes requested):
I think the settings page definitely needs some nicer look to it, and I'm not sure if the `OverviewContent` will fit to that. But I don't think it's worth speculating too much now.

The functionality seems solid 👍

IidaKainu:
> Could select by id directly from the query, and skip when no id is present.


Neat, added

juhoeerola (approved):

PR #422, opened 2022-05-18
Add sanity check and better typing for paths

juhoeerola:
It hasn't been once or twice when I've had to peek into the definition
to see if I should add a separator or not.

This enforces a consistent format for the defined segments:
- segment must start with the separator
- segment must not end with the separator, unless it consist only of the
  separator

The format as a regular expression:
/|/.*[^/]

This also provides literal types for the segments, making it easier to
look up the value when necessary.

Also removed one unused duplicate definition.


IidaKainu (approved):
👍

PR #423, opened 2022-05-18
Add task details tooltip to MilestonesEditor

IidaKainu:
Tooltip shows the avg. task complexity and value and avg. value for each customer.


frontend/src/components/SortableTask.module.scss
```diff
@@ -34,3 +34,23 @@
...
+    margin-bottom: 2px;
+  }
+
+  .customerValues {
+    @extend .layout-row;
+    gap: 15px;
+
+    & > div {
+      @extend .layout-row;
+      gap: 4.75px;
```
juhoeerola, line 53: I'm not sure if it's just me, but splitting pixels feels a bit odd. Could we manage with a whole number?

IidaKainu: The value is probably from Figma, but could be rounded.

frontend/src/components/SortableTask.module.scss
```diff
@@ -34,3 +34,18 @@
...
+  @extend .layout-column;
+  align-items: center;
+  gap: 20px;
+
+  .title {
+    margin-bottom: 2px;
+  }
+
+  .customerValues {
+    @extend .layout-row;
```
juhoeerola, line 48:
```suggestion
    display: grid;
    grid-template-columns: repeat(3, auto);
```

The flex row overflows with some more customers.

Maybe could use a grid with some number of columns instead, allowing the content to flow to multiple rows.

juhoeerola (changes requested):
The tooltip doesn't scale well above 4 clients.

Otherwise looks good

frontend/src/components/SortableTask.module.scss
IidaKainu: Good point. Implemented with flex-wrap (it is easier to center to me)

juhoeerola: That works as well, if not better 👍

juhoeerola (approved):

PR #424, opened 2022-05-18
Add default sort prop to RatingTable

IidaKainu:
Default sorting for complexity-table is creator instead of the customer.
Only solves part of the problem: customer is still selectable as sorting, which is useless for complexity-table. I made [a card about that](https://trello.com/c/CopQGUnQ/696-taskoverview-poista-customer-sort-vaihtoehto-complexity-rating-tablesta).


PatricKangasmaki (approved):
Great way to do the change 👍

PR #426, opened 2022-05-20
Correct TaskMap portal creation

IidaKainu:
The portal containing dragged tasks was attached to TaskMapPage, where zoom level was not in scope. This lead to broken dragging styles.

After the correction, most dragging styles are fixed. However, Handle's border-width remains broken for unknown reasons.


juhoeerola:
> Handle's border-width remains broken for unknown reasons.

I couldn't see this, it looked like it was all working fine for me 🤔

IidaKainu:
> I couldn't see this, it looked like it was all working fine for me 🤔

I could not replicate this either today. Maybe it had something to do with caching.

juhoeerola (approved):

PR #428, opened 2022-05-24
Add context menu for removing taskmap edges

PatricKangasmaki:
Right or left clicking edges on task map now opens context menu where you can delete the relation.

### DIFF
--- a/frontend/src/components/TaskMapEdge.tsx
+++ b/frontend/src/components/TaskMapEdge.tsx
@@ -7,6 +7,9 @@ import {
   Position,
   MarkerType,
 } from 'react-flow-renderer';
+import Menu from '@mui/material/Menu';
+import MenuItem from '@mui/material/MenuItem';
+import DeleteIcon from '@mui/icons-material/Delete';
 import { chosenRoadmapIdSelector } from '../redux/roadmaps/selectors';
 import css from './TaskMapEdge.module.scss';
 import { apiV2 } from '../api/api';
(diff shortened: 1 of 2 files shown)


frontend/src/components/TaskMapEdge.tsx
```diff
@@ -96,13 +117,47 @@ const DrawPath: FC<DrawPathProps> = ({
...
+      <Menu
+        PaperProps={{
+          style: {
+            display: 'flex',
+            alignItems: 'center',
+            overflow: 'hidden',
+            height: '30px',
+            borderRadius: '20px',
+          },
+        }}
```
PatricKangasmaki, line 139: I tried giving Menu both `className` and `classes`. Both of them worked (some of the styles applied from css file, such as border-radius) but they broke the styles for some reason and everything looked different compared to inline styles. For example, height and the width of the menu worked really inconsistently. I couldn't find a fix for this, do you have any suggestions or is this inline styling fine here?

IidaKainu: Finally seemed to found a solution to our mui styling woes. From https://mui.com/material-ui/guides/interoperability/#css-modules:
> It's important to keep in mind that CSS Modules adds an unique id to each class, and that id won't be present on the MUI provided children class. To escape from that, CSS Modules provides a functionality, the :global selector.

Mui-styling in non-module scss file has worked, because a unique id is not added. However, that would mean moving the styles to shared.scss or creating a new file, which is not desirable in all cases. To avoid this, I added a `className={classes(css.menu)}` to the menu and :global in the css-selector, which works. Should be useful in the future too!
```
.menu :global .MuiMenu-paper {
  display: flex;
  align-items: center;
  overflow: hidden;
  height: 30px;
  border-radius: 20px;
}
```

IidaKainu (changes requested):

PatricKangasmaki:
Added all the suggestions. Thanks for the help on the styles, that's a really good solution 🙏

IidaKainu (approved):
💯

PR #429, opened 2022-05-24
Add tooltip to the taskmap position reset -button

PatricKangasmaki:
Added a tooltip on the task group reset button on the task map so it's a bit more clear to the user


IidaKainu (approved):

IidaKainu:
Just noticed the fit view -button has its own tooltip. Can it be changed to ours for consistency?
image

PR #430, opened 2022-05-24
Move taskmap controls to top corner

PatricKangasmaki:
Moved the task map controls from bottom left to top right corner. Also made the icons a bit larger, any opinions on that? Are they good now or should they be changed more (to either smaller or bigger)

### DIFF
--- a/frontend/src/components/TaskMap.module.scss
+++ b/frontend/src/components/TaskMap.module.scss
@@ -7,17 +7,26 @@ $ZOOM: var(--zoom, 1);
   width: 100%;
 }
 
-.info {
-  position: relative;
-  left: 48px;
-  bottom: 26px;
-  width: 16px;
+.controlsContainer {
+  display: flex;
+  justify-content: flex-end;
+  margin-right: 80px;
 }
 
-.restart {
+.controls {
+  display: flex;
+  align-items: center;
   position: relative;
+  box-shadow: none;
+  top: 10px;
+  gap: 7px;
+}
+
+.controlButton {
   cursor: pointer;
-  left: 28px;
-  bottom: 49.5px;
-  width: 16px;
+}
+
+.taskmapInfo {
+  width: 27px;
+  height: 27px;
 }
--- a/frontend/src/components/TaskMap.tsx
+++ b/frontend/src/components/TaskMap.tsx
@@ -11,6 +11,7 @@ import ReactFlow, {
 } from 'react-flow-renderer';
 import classNames from 'classnames';
 import InfoIcon from '@mui/icons-material/InfoOutlined';
+import FullscreenIcon from '@mui/icons-material/Fullscreen';
 import { useTranslation } from 'react-i18next';
 import RestartAltIcon from '@mui/icons-material/RestartAlt';
 import {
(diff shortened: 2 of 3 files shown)


PatricKangasmaki:
Removed default `showFitView` from React flow controls and replaced it with MUI icon with fitView() onClick func. Added a tooltip for it as well

frontend/src/components/TaskMap.module.scss
```diff
@@ -7,17 +7,26 @@ $ZOOM: var(--zoom, 1);
...
-  width: 16px;
+.controlsContainer {
+  display: flex;
+  justify-content: flex-end;
+  margin-right: 80px;
 }
 
-.restart {
+.controls {
+  display: flex;
```
IidaKainu, line 17: The controls look slightly better to me centered
```suggestion
  display: flex;
  align-items: center;
```

frontend/src/components/TaskMap.tsx
```diff
@@ -279,19 +280,38 @@ export const TaskMap: FC<{
...
-            />
-          </InfoTooltip>
-        </Controls>
+        <div className={classes(css.controlsContainer)}>
+          <Controls
+            className={classes(css.controls)}
+            showInteractive={false}
+            showZoom={false}
+            showFitView={false}
+          >
```
IidaKainu, line 289: Do we need the Controls component anymore as each control is false? 
Without this, there is no need for relative control styles (example below)

```
.controlsContainer {
  @extend .layout-row;
  justify-content: flex-end;
  align-items: center;
  margin-right: 10px;
  margin-top: 10px;
  gap: 7px;
}

.controlButton {
  cursor: pointer;
}
```

frontend/src/i18/english.ts
```diff
@@ -419,6 +419,7 @@ export const english = {
     'Client shares score': 'Client shares score',
     Poor: 'Poor',
     Great: 'Great',
-    'Reset taskgroup positions': 'Reset taskgroup positions',
+    'Fit view -tooltip': 'Fit taskgroups to the current viewport',
+    'Taskgroup reset -tooltip': 'Reset task group positions',
```
IidaKainu, line 423: The new tooltip msg is a little long, could the following work?
```suggestion
    'Fit view -tooltip': 'Fit groups to the viewport',
    'Taskgroup reset -tooltip': 'Reset group positions',
```

IidaKainu (changes requested):
The new buttons look much more noticeable, great progress 👍

frontend/src/components/TaskMap.tsx
PatricKangasmaki: I also tried removing this as I noticed that I disabled all the functionalities but the buttons don't seem to work on the canvas unless they're inside the <Controls> .. 😅

frontend/src/i18/english.ts
PatricKangasmaki: Yeah these seem better! I'll do the change

frontend/src/components/TaskMap.module.scss
PatricKangasmaki: This seems good. I also changed the infoIcon to be 1px larger, so I could get rid of the `marginTop: 2px` at the control buttons - now they look nicely aligned

PatricKangasmaki:
Added your suggestions 👍

IidaKainu (approved):

PR #431, opened 2022-05-24
Add navigation from task overview to planner

IidaKainu:
If user has navigated to TaskOverview from RelationsModal link, the back-button takes the user back to MilestonesEditor (with the same RelationsModal open)

### DIFF
--- a/frontend/src/redux/roadmaps/types.ts
+++ b/frontend/src/redux/roadmaps/types.ts
@@ -8,7 +8,7 @@ import {
 export interface RoadmapsState {
   selectedRoadmapId?: number | null;
   taskmapPosition: TaskmapPosition | undefined;
-  fromMilestonesEditor?: boolean;
+  fromMilestonesEditor?: string;
 }
 
 export interface Customer {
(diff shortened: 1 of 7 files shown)


IidaKainu:
Though, should the (same) RelationsModal be open when navigating back to MilestonesEditor? I could make that happen

frontend/src/redux/roadmaps/types.ts
```diff
@@ -8,6 +8,7 @@ import {
 export interface RoadmapsState {
   selectedRoadmapId?: number | null;
   taskmapPosition: TaskmapPosition | undefined;
+  fromMilestonesEditor?: string;
```
juhoeerola, line 11: I think the more relevant information is that this is the "back href for task overview page", which should be in the name. Currently the name doesn't really tell much, unless you already know what it is supposed to mean.

This is even more apparent when handling the other locations.

juhoeerola (changes requested):
There are links to the task overview from elsewhere also:
- the individual tasks in the relation "modal"
- the tasks in roadmap tab sidebar (this won't stay open with the current implementation)

The `TaskRow` (in TaskTable.tsx) component links to the task overview and could also set the back ref (when it's not on the task overview page already).

---

It would be neat if we could use the history stack for this (instead of extra redux state), but that would be a bit tricky when browsing between the tasks using the arrows on the overview page.

IidaKainu:
Took your feedback into consideration and [implemented this with another strategy](https://github.com/Vincit/VISDOM-Roadmapper/pull/439).

PR #434, opened 2022-05-24
Roadmap sidebar metric texts

AnttiVaarala:


frontend/src/components/RoadmapGraphSidebar.module.scss
```diff
@@ -30,9 +30,18 @@
 .metrics {
   @extend .layout-row;
   gap: 10px;
-  height: 120px;
+  height: max-content;
   align-items: center;
   justify-content: center;
+  div {
+    white-space: normal !important;
+    height: calc(100% - 8px);
```
IidaKainu, line 38: This is not a likely scenario, but the "dash" can become misshapen if the value has spaces.
To avoid this, the border on MetricsSummary .dash could be changed to `border-top: 1px solid $COLOR_BLACK40;`
image

IidaKainu (approved):
The metrics summary looks great 👍 
Thoughts not necessarily related to RoadmapGraphSidebar below (but can be added in this pr if you agree on them)

PR #436, opened 2022-05-25
Hide completed tasks from TaskHeatmap

IidaKainu:
https://trello.com/c/FuntYSh9/624-dashboard-valmiit-taskit-pois-heatmapista

### DIFF
--- a/frontend/src/utils/TaskUtils.ts
+++ b/frontend/src/utils/TaskUtils.ts
@@ -150,10 +150,11 @@ export const ratedByCustomer = (
       rating.forCustomer === customer.id && rating.createdByUser === rep.id,
   );
 
-export const completed = (task: Task) => task.status === TaskStatus.COMPLETED;
+export const isCompletedTask = (task: Task) =>
+  task.status === TaskStatus.COMPLETED;
 
 export const isCompletedMilestone = ({ tasks }: Version) =>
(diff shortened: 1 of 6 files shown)


frontend/src/utils/TaskUtils.ts
```diff
@@ -150,7 +150,7 @@ export const ratedByCustomer = (
       rating.forCustomer === customer.id && rating.createdByUser === rep.id,
   );
 
-const completed = (task: Task) => task.status === TaskStatus.COMPLETED;
+export const completed = (task: Task) => task.status === TaskStatus.COMPLETED;
```
juhoeerola, line 153: Now that this is exported, there would be couple of places to use it. I changed those to use this.

Also, this was named in the single file context, not considering exporting.
I'm not sure if it's the best name (as exported), but I think it's good enough together with the function signature.

juhoeerola (approved):
👍 

If you have a perfect name in mind for `completed()`, could *maybe* consider renaming.

Check my commit before merging, and keep or drop it.

IidaKainu:
> Check my commit before merging, and keep or drop it.

Looks like a great addition, thanks 👍 

> If you have a perfect name in mind for completed(), could maybe consider renaming.

We do have isCompletedMilestone-function, so the natural name could be isCompletedTask even though it is a little longer. I'll change it and merge

PR #437, opened 2022-05-25
Fix landingpage card hr style

AnttiVaarala:


IidaKainu (approved):
Fixes the empty space in the line 👍 
Do still consider changing the color back to $COLOR_AZURE, since the front page has a subtle design theme of blue lines and we could be breaking the intent here

PR #438, opened 2022-05-25
Task overview editable select style

AnttiVaarala:
task overview editable select on same line + textarea style fixes

### DIFF
--- a/frontend/src/components/EditableSelect.module.scss
+++ b/frontend/src/components/EditableSelect.module.scss
@@ -9,6 +9,15 @@
   margin-top: 0;
 }
 
+.editButtonDiv {
+  margin-top: 0;
+  text-align: start;
+}
+
+.value {
+  margin: 0;
+  flex-grow: 1;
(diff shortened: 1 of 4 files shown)


frontend/src/components/EditableSelect.module.scss
```diff
@@ -9,6 +9,15 @@
   margin-top: 0;
 }
 
+.editButtonDiv {
+  margin-top: 0;
```
IidaKainu, line 13: I did not see this affecting the styles (?). If that's correct, the line can be removed

IidaKainu (approved):
Looks good to go, but consider the comment below before merging 👍

juhoeerola:
What is the status of this one?

The same changes seem to also be in #444

AnttiVaarala:
It's the same, this was originally just for the editableSelect styling but there were many other changes for task overview so I added all the task overview changes to same branch, I'll close this

PR #439, opened 2022-05-26
Taskoverview/navigation

IidaKainu:
Use location state to use a custom url in taskoverview's back-button.
- React Router does not grant direct access to the history stack, since its mutable. Instead it has [location state](https://v5.reactrouter.com/web/api/location), which seems like the correct place to accomplish this.
- When navigating to taskoverview, pass an object `{ pathname: string, state: { from: string; } }` instead of the task overview url. Then the back-button will redirect to the `from` property url

This is used in TaskRow (except when location is already task overview) and the relation "modal"

RoadmapGraphSidebar is changed to VersionDetailsModal. This is to get a modal url, so we can navigate back to it from task overview.


frontend/src/pages/TaskOverviewPage.tsx
```diff
@@ -164,12 +164,18 @@ export const getTaskOverviewData = (
...
+}
+
 const TaskOverview: FC<{
   tasks: Task[];
   task: Task;
   taskIdx: number;
 }> = ({ tasks, task, taskIdx }) => {
   const history = useHistory();
+  const location = useLocation<LocationState>();
+  const fromUrl = (location.state as LocationState | undefined)?.from;
```
juhoeerola, line 178:
```suggestion
  const { state } = useLocation<LocationState | undefined>();
```

This could be simplified a bit.

And `undefined` can be added to the type to avoid the typecast later.

frontend/src/pages/TaskOverviewPage.tsx
```diff
@@ -221,11 +227,16 @@ const TaskOverview: FC<{
   return (
     <div className="overviewContainer">
       <Overview
-        backHref={`${tasksPage}${paths.tasksRelative.tasklist}`}
+        backHref={fromUrl || `${tasksPage}${paths.tasksRelative.tasklist}`}
```
juhoeerola, line 230:
```suggestion
        backHref={state?.from || `${tasksPage}${paths.tasksRelative.tasklist}`}
```

frontend/src/pages/TaskOverviewPage.tsx
```diff
@@ -221,11 +227,16 @@ const TaskOverview: FC<{
...
         overviewType={t('Task')}
         name={task.name}
         previousAndNext={siblingTasks}
-        onOverviewChange={(id) => history.push(`${tasksPage}/${id}`)}
+        onOverviewChange={(id) =>
+          history.push({
+            pathname: `${tasksPage}/${id}`,
+            state: { from: fromUrl },
+          })
```
juhoeerola, line 238:
```suggestion
          history.push({
            pathname: `${tasksPage}/${id}`,
            state,
          })
```

juhoeerola (changes requested):
This seems quite clean implementation 👍 

One minor bug, and couple suggestions.

IidaKainu:
Great points - added fixes 👍

juhoeerola (approved):

PR #441, opened 2022-05-30
Rating value sort highest/lowest value

juhoeerola:
Also removed uses of deprecated `String.prototype.substr()`


IidaKainu (approved):
Neat implementation 👍

PR #442, opened 2022-05-30
Add a gap to client name and dot in client details

IidaKainu:
Small fix: https://trello.com/c/UZEssfWr/701-clientoverview-pallon-ja-nimen-v%C3%A4lille-lis%C3%A4%C3%A4-v%C3%A4li%C3%A4

### DIFF
--- a/frontend/src/pages/ClientOverviewPage.module.scss
+++ b/frontend/src/pages/ClientOverviewPage.module.scss
@@ -4,3 +4,9 @@
   min-width: 400px;
   width: 20%;
 }
+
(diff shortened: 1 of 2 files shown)


juhoeerola (approved):
Seems good 👍 

This was implemented twice, but I think we can should go with this since this is
- bit more focused
- also centres the dot vertically
- opened earlier
- also in trello

 closes #445

juhoeerola:
closes #445

(apparently github doesn't notice mentions in review comments...)

PR #443, opened 2022-05-30
Consider empty milestones in TimeEstimationPage

IidaKainu:
Empty milestones are excluded from the dropdown. 
If all milestones are empty, show a empty dropdown with "No milestones available" message.


juhoeerola (approved):
👍

PR #444, opened 2022-05-30
Fix task overview style

AnttiVaarala:


frontend/src/components/Overview.tsx
```diff
@@ -61,20 +61,25 @@ export const OverviewContent: FC<OverviewContentProps> = ({
...
-                </div>
-              ) : (
-                <div className={classes(css.value, css[format!])}>{value}</div>
-              )}
-            </div>
-          ))}
-        </div>
+        <>
+          <div className={classes(css.separator)} />
+          <div className={classes(css.column)} key={idx}>
```
juhoeerola, line 66:
```suggestion
        <div key={idx}>
          <div className={classes(css.separator)} />
          <div className={classes(css.column)}>
```

The `key` is now in an incorrect place. Just changing the fragment into a `div` with the key doesn't seem to work with the styles though.

juhoeerola (changes requested):

frontend/src/components/Overview.tsx
juhoeerola, line 66:
```suggestion
          <div className={classes(css.column)} key={idx}>
```

It might be even simpler to move the border to the `.column` class, something like:

```scss
  .column {
    @extend .layout-column;
    border-left: 1px solid $COLOR_BLACK40;
    padding-left: 10px;
    min-height: 50px;
```

(Sorry for the spam, slow brain today.)

IidaKainu:
Can you rebase this on master? 🙏 I added vertical overview styles - I could go check how the borders look there and disable them in vertical mode if needed

frontend/src/components/Overview.tsx
```diff
@@ -60,8 +60,14 @@ export const OverviewContent: FC<OverviewContentProps> = ({
     )}
     <div className={classes(css.data)}>
       {data.map((column, idx) => (
-        // eslint-disable-next-line react/no-array-index-key
-        <div className={classes(css.column)} key={idx}>
+        <div
+          className={classes(
+            css.column,
+            `${idx + 1 === data.length && css.lastColumn}`,
+          )}
```
juhoeerola, line 69:
```suggestion
          className={classes(css.column, {
            [css.borderless]: idx + 1 === data.length || vertical,
          })}
```

This is the better way to add conditional class names. The current one would break as soon as we add a `.false` class (I hope we won't, but still...).

Also, the border seems a bit odd with the vertical layout, so maybe disable it then as well.
Maybe the class could be renamed, as "lastColumn" wouldn't be accurate anymore.

AnttiVaarala:
It's rebased now!

frontend/src/components/Overview.module.scss
```diff
@@ -34,10 +34,11 @@
...
     textarea {
-      overflow: scroll !important;
+      overflow: auto !important;
+      background-color: $COLOR_WHITE;
     }
     div:nth-child(2) {
-      padding-right: 15%;
+      padding-right: 5%;
       justify-content: left;
     }
```
juhoeerola, line 43: What is this supposed to target?
Whatever it is, I think this is matching some unwanted elements as well.

I'd also prefer a fixed (`px`, `rem`, etc.) padding instead of relative (`%`), especially as the targeted element isn't clear.

frontend/src/components/EditableSelect.module.scss
```diff
@@ -9,6 +9,15 @@
...
+.editButtonDiv {
+  text-align: end;
+  width: min-content;
+}
+
+.value {
+  margin: 0;
+  flex-grow: 1;
+}
 .selectWrapper {
```
juhoeerola, line 21:
```suggestion
}

.selectWrapper {
```

nitpick: I'd add a line between toplevel classes

juhoeerola (changes requested):

frontend/src/components/Overview.module.scss
AnttiVaarala: It targets the edit button so there's gap after the edit button but now I realized I could just give padding-right: 10px; for the .editableContainer and it does the job

frontend/src/components/Overview.module.scss
```diff
@@ -32,13 +32,11 @@
     overflow-wrap: anywhere;
     align-items: start;
   }
+  padding-right: 10px;
```
juhoeerola, line 35: nice improvement 👍

juhoeerola (approved):
Looking good

PR #446, opened 2022-06-01
Overview - fix scroll on update

ghost:
The reason for the scroll was that the affected components rendered a `<LoadingSpinner />` whenever the data was refetched. This caused major layout change which was seen as scrolling.

Now the components only show the spinner on the initial load, refetching is done in the background since we can show the old data in the meantime.

---

Additionally fixed couple issues with stale data being shown:
- customer changes did not reflect in `User.representativeFor`
- taskmap overview didn't reflect changes in `tasks`

---

Also removed redundant typing from jsdoc comment.

### DIFF
--- a/frontend/src/components/Overview.tsx
+++ b/frontend/src/components/Overview.tsx
@@ -82,18 +82,18 @@ export const OverviewContent: FC<OverviewContentProps> = ({
 
 /**
  * Renders a <Overview /> component, which can be used in task, client etc. overviews.
- * @param {object} props Component props
- * @param {string} props.backHref Href for the back button's link
- * @param {string} props.overviewType Type of the overview. Appears in the header text
- * @param {any} props.name Name of the overviewed object. Appears in the header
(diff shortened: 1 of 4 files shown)


frontend/src/components/Overview.tsx
```diff
@@ -82,18 +82,18 @@ export const OverviewContent: FC<OverviewContentProps> = ({
 
 /**
  * Renders a <Overview /> component, which can be used in task, client etc. overviews.
- * @param {object} props Component props
- * @param {string} props.backHref Href for the back button's link
```
IidaKainu, line 86: Good change. At the time of writing this I was unfamiliar how JSDoc should be written for React & typescript, but thinking back the type annotations are probably unnecessary.

IidaKainu (approved):
🔥 
I expected fixing the scroll issue would take much more effort, but great to see you've found a simple solution. The additional fixes look good to me as well.

PR #447, opened 2022-06-01
Test data from demo sessions

juhoeerola:


server/package.json
```diff
@@ -21,7 +21,7 @@
   "scripts": {
     "migrate": "knex --knexfile=./knexfile.ts migrate:latest",
     "seed-db": "knex --knexfile=./knexfile.ts seed:run",
-    "start": "knex --knexfile=./knexfile.ts migrate:latest && knex --knexfile=./knexfile.ts seed:run && ts-node ./src/server.ts",
+    "start": "ts-node ./src/server.ts",
```
IidaKainu, line 24: Is there a reason to remove the migration from the start script? If not, it would be neat to include it (because I will surely forgot to run `yarn migrate` in the beginning)

juhoeerola: The reason I took it out was that the seed-db was in between.

Although this could be added before both seed-db and start, since there is no harm done trying to migrate multiple times.

IidaKainu (commented):
I looked through the data in sql console and confirmed there was no extra/specific data included.
The Trello external links are left in the data - is it our Roadmapper Trello board or some demo board? If the latter, maybe we should remove them from the data.

juhoeerola:
> is it our Roadmapper Trello board or some demo board?

Doesn't seem to be our trello, probably some demo board then.

I could make all the task "internal" (not imported)

PR #448, opened 2022-06-02
Change rating sort option wording

juhoeerola:


frontend/src/components/RatingTable.tsx
```diff
@@ -138,8 +138,9 @@ export const ratingTable: (def: RatingTableDef) => FC<RatingTableProps> = ({
           <Trans i18nKey="List order" />
           <Dropdown
             css={css}
-            title={t(TaskRatingSortingTypesToText(sorting.type.get()))}
-            maxLength={20}
```
juhoeerola, line 142: "complexity" didn't fit the limit, and since the possible title values are fixed, I removed the limit all together instead of guessing a better limit.

frontend/src/i18/english.ts
```diff
@@ -71,8 +71,8 @@ export const english = {
     'Created on': 'Created on',
     Name: 'Name',
     Value: 'Value',
-    'Value ascending': 'Value, lowest first',
-    'Value descending': 'Value, highest first',
+    'Value ascending': 'Lowest {{type}} first',
+    'Value descending': 'Highest {{type}} first',
```
IidaKainu, line 75: The titles get a little clipped due to lack of space (a horizontal scrollbar does appear though): would only "Lowest/Highest {{type}}" suffice?
image

IidaKainu (changes requested):

frontend/src/i18/english.ts
juhoeerola: How about anchoring the menu from the right side?

That keeps the menu visible, and allows room for the more complete description for the ordering.

IidaKainu (approved):
> How about anchoring the menu from the right side?

Nice solution 👍 Other improvements also look good

PR #449, opened 2022-06-02
Fix having to refresh for sockets to work

JimiRantanen:
Websocket refresh only worked if you refreshed the page while you had a roadmap selected - socket communication relies on users being added in rooms corresponding to their userId when they connect, and the server was making them leave those rooms when no roadmap was selected.


juhoeerola (approved):
Great

PR #450, opened 2022-06-02
Show info for unordered tasks

juhoeerola:
Showing both felt quite cramped, so I just left out the ratings as those are also found in the tooltip.

The disadvantage is that it's no longer easy to look through all the tasks in the list, as one must open the tooltip first.
For that some indication of the important tasks could be useful, so the user knows which tooltips to check.


IidaKainu (approved):
> Showing both felt quite cramped, so I just left out the ratings as those are also found in the tooltip.
> 
> The disadvantage is that it's no longer easy to look through all the tasks in the list, as one must open the tooltip first.

I agree with your decision to leave the ratings out. We do have a problem with planner requiring a lot of memorization though.

> For that some indication of the important tasks could be useful, so the user knows which tooltips to check.

Could be a step to the right direction. Let's bring the issue up with our designer.

PR #451, opened 2022-06-03
Remove hover effect from taskmap edge context menu

PatricKangasmaki:
Now the context menu opens fully when clicking the edges instead of having the opening hover-effect


frontend/src/components/TaskMapEdge.module.scss
```diff
@@ -45,24 +45,14 @@
   background-color: $COLOR_WHITE !important; // !important to override MUI default styles
   color: $COLOR_BLACK10; // Adds color for default effect from MUI: turn background color of the menu item to lower shade of the "color" on click
   width: 48px;
-  &:hover {
-    transition: width 0.4s;
-    width: 185px;
-  }
+  transition: width 0.4s;
```
IidaKainu, line 48: I don't think affects anything currently(?)

frontend/src/components/TaskMapEdge.module.scss
```diff
@@ -45,24 +45,14 @@
   background-color: $COLOR_WHITE !important; // !important to override MUI default styles
   color: $COLOR_BLACK10; // Adds color for default effect from MUI: turn background color of the menu item to lower shade of the "color" on click
   width: 48px;
```
IidaKainu, line 47: Looks like this is an old width (there's another few lines below)
```suggestion
```

IidaKainu (changes requested):
Noticed few leftovers, otherwise looking good

PatricKangasmaki:
Good catches, removed them both 👍

frontend/src/components/TaskMapEdge.module.scss
```diff
@@ -44,25 +44,13 @@
 .menuItem {
   background-color: $COLOR_WHITE !important; // !important to override MUI default styles
   color: $COLOR_BLACK10; // Adds color for default effect from MUI: turn background color of the menu item to lower shade of the "color" on click
-  width: 48px;
-  &:hover {
-    transition: width 0.4s;
-    width: 185px;
-  }
+  width: 185px;
```
IidaKainu, line 47: Other (if we had some) translations could take up more space, so changing to max-content would be a better practice
On the other hand, I did not see removing width changing anything, so is it required at all?
```suggestion
  width: max-content;
```

IidaKainu (approved):
The following suggestion is so small for another change-request-cycle that I'd rather approve this now. But do consider the suggestion (and I'm sure your solution is 👍)

frontend/src/components/TaskMapEdge.module.scss
PatricKangasmaki: Ah, that's a good point. Width seems to be defaulted at max-content so removing it altogether should be good. I'll do the change and add translation for the "remove relation" as well, seems like it was missing 👍

IidaKainu (approved):
Double approved: good catch on the translations

PR #452, opened 2022-06-03
Add vertical Overview styles

IidaKainu:
Use vertical styling in settings-page project overview

### DIFF
--- a/frontend/src/components/Overview.module.scss
+++ b/frontend/src/components/Overview.module.scss
@@ -63,6 +63,11 @@
   @extend .layout-row;
   width: 100%;
   gap: 5%;
+
+  &.vertical,
+  &.vertical .data {
+    flex-direction: column;
+  }
 }
 
 .metrics {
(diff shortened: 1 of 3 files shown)


frontend/src/components/Overview.module.scss
```diff
@@ -63,6 +63,13 @@
   @extend .layout-row;
   width: 100%;
   gap: 5%;
+
+  &.vertical {
+    @extend .layout-column;
+    .data {
+      @extend .layout-column;
+    }
+  }
```
PatricKangasmaki, line 72: This bothers me a bit as it translates to:

```
&.vertical {  
    display: flex;
    flex-direction: column;
    .data {
       display: flex;
       flex-direction: column;
    }
  }
```

Could this be achieved without repeating the code? For example this would look the same but is there some reason to have .vertical and .data defined separately?

```
  &.vertical .data {
    @extend .layout-column;
  }
```

IidaKainu: I also want to apply the `flex-direction: column` to .content&.vertical, but I'll clean this up 👍

PatricKangasmaki (approved):
Looking good 👍

PR #453, opened 2022-06-03
Fix topbar shrinking behaviour

AnttiVaarala:
TopBar was shrinking too much with (really) small screen width and contents of the TopBar overflowed, setting min-width: min-content for .topBar fixes the issue


IidaKainu (approved):
Nice 👍 fixes the problem

PR #454, opened 2022-06-03
Fix graph sizing bug

juhoeerola:
The items would not be displayed at all if all the values were the same.
Thus a single milestone would never be shown.


IidaKainu (approved):
Great job on catching this bug 👍

PR #455, opened 2022-06-06
Add explanations to relation on task overview

juhoeerola:


IidaKainu (approved):
Looks good

With the addition, the different colored border "lines" started to bother me. Added a commit to fix this; include if you think it is an improvement.

juhoeerola:
I think black 10 is bit low contrast for the thin 1px lines, but I guess it depends how subtle the lines are meant to be.

At the moment the default seems to come from bootstrap:
```css
hr {
    margin-top: 1rem;
    margin-bottom: 1rem;
    border: 0;
    border-top: 1px solid rgba(0,0,0,.1);
}
```

juhoeerola:
Looking at figma, it seems that black 10 is used for the lines, so let's go with that.

PR #457, opened 2022-06-06
Add weighted summary section

juhoeerola:
I think the milestone total/average calculations should be revisited.

But with the current calculations, this should do.


IidaKainu (commented):
The changes look good. Just have not approved yet, because I am wondering how the upcoming value calculation change would affect what metrics we need to show in milestones

PR #458, opened 2022-06-07
Integration styles

IidaKainu:
"Fix" integration styles according to the design
- due to other content in settings page, the design was only loosely followed
  - For example, h2 was already taken up in "project" and "integrations". Due to this, smaller typographies were chosen for headers
- our dropdown component is used in the design, so some improvements were also made to them


IidaKainu:
The commit [Add id prop for the Dropdown](https://github.com/Vincit/VISDOM-Roadmapper/pull/458/commits/16e6f410382389f2abfeeedc0ae095df36d636e2) is a little questionable. 
- IntegrationConfiguration contained two label-elements with a htmlFor-attibutes. From what I tried to google, I am not completely sure can label be connected to a button (our dropdown). I suppose it can, but usually does not make sense as buttons can have an aria-label prop. 

Any opinions on whether to keep it as is or **remove the htmlFor-attribute from labels and add an aria-label to the button**?

frontend/src/components/forms/Dropdown.tsx
```diff
@@ -64,12 +67,13 @@ export const Dropdown: FC<{
...
 
   return (
-    <div className={classes(css.dropContainer)}>
+    <div className={classes(css.dropdownContainer)}>
       <button
         type="button"
         className={classes(css.dropButton)}
         onClick={showMenu}
         disabled={disabled}
+        id={id}
```
juhoeerola, line 76: Now that the `id` is added to this, the dropdown can be opened by clicking the corresponding label.

That's totally fine, but now the dropdown can be closed by clicking practically anywhere on the page *except the label* (which is a bit odd user experience). This could be avoided by preventing the default on the closing click handler.

juhoeerola (changes requested):
One bug waiting to happen, one UX oddity

and I will sure miss the searchability of the select component e.g. for the task state selection.

Also the disabled dropdown is missing the left margin, and the text may go under the arrow icon.

Screenshot 2022-06-07 at 15 40 02

IidaKainu:
Added the review fixes: thanks for pointing them out

Additionally: 
- "not started" can't be selected as a task state. A tooltip is added to "Relations" header informing tasks are "not started" by default. 
- Added the non-selected dropdown style to newly selected columns (picture below). A placeholder prop is added to the dropdown to change the title to "not selected"
image

IidaKainu:
> The "X board" and "task state in X" dropdowns' text may overflow, it should be limited either by the max length prop or some css.

Good catch. I don't see a reason to limit the length here, so changed the css to fit the title content

juhoeerola (approved):
Seems good 👍 

Should we add the separator label for the staged and submitted mappings in this PR?

IidaKainu:
> Should we add the separator label for the staged and submitted mappings in this PR?

@JimiRantanen is probably working on it. Let's create another pr for that

PR #460, opened 2022-06-07
Add margin between multiple scrollbars

IidaKainu:
https://trello.com/c/S5O0WEsp/705-marginia-scollbareihin
I would have added space with margins, but they seemed to have no effect on the scrollbar. So implemented this with borders instead.

### DIFF
--- a/frontend/src/pages/ProjectOverviewPage.module.scss
+++ b/frontend/src/pages/ProjectOverviewPage.module.scss
@@ -5,9 +5,8 @@
   flex-direction: column;
   align-items: flex-start;
   padding-top: 100px;
-  height: 100%;
+  flex-grow: 1;
   width: 100%;
-  padding-right: 2px;
   padding-bottom: 2px;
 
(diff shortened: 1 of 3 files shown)


IidaKainu:
Forgot horizontal scrollbars exist. Now border is added to all sides

juhoeerola:
I think it would be better to add small padding to the scrollable container in the places where the bars would be next to each other.

The appearance becomes a bit wild with the border as there's now 3 shades of gray in the bar (handle, background, border), and the page background is likely one more shade.

juhoeerola (approved):
Seems good 👍 

The overview's duplicate vertical scroll didn't make much sense, so I removed that. Check that I didn't mess it up somehow.

IidaKainu:
> The overview's duplicate vertical scroll didn't make much sense, so I removed that. Check that I didn't mess it up somehow.

Great! horizontal scroll still works as usual so should be ok

PR #461, opened 2022-06-07
Remove unnecessary scrollbars on milestone editor page

juhoeerola:
Now the topbar section is independently scrollable and there should be only one horizontal bar at the bottom.

Also added larger minimum height for the milestones, as the content
would overflow on the previous minimum.

### DIFF
--- a/frontend/src/pages/MilestonesEditor.module.scss
+++ b/frontend/src/pages/MilestonesEditor.module.scss
@@ -144,9 +144,9 @@
 }
 
 .topbar {
+  @extend .layout-row;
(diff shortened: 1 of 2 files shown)


IidaKainu (changes requested):
Looks quite good. However, the topbar now has a vertical scrollbar: could it take up all the vertical space needed instead?
image

IidaKainu (approved):
> Even now it will get a scrollbar once the page is so small you can't reasonably do anything with it.

Looks better 👍 The page does need to be really small for that, so let's go with the current implementation

PR #464, opened 2022-06-08
Stop propagating clicks from tooltip popper

juhoeerola:
Previously the click event would propagate to the parent element of the
tooltip, which would be unexpected and undesirable in most cases.


IidaKainu (approved):
🌲/6

PR #465, opened 2022-06-08
Hide overflowing "waiting for ratings" dots

juhoeerola:
The column has room for minimum of 6 dots.

Showing the number of hidden dots might be "nice to have" option, but I'm not sure if that would be reasonable to implement.


JimiRantanen (approved):
Works as advertised

PR #467, opened 2022-06-08
Fix drag issue with nested scroll containers

JimiRantanen:


frontend/src/components/SortableTaskList.tsx
```diff
@@ -41,15 +47,21 @@ export const SortableTaskList: FC<{
...
 
   useEffect(() => {
     setSearched(
       tasks.filter(({ name }) => name.toLowerCase().includes(search)),
     );
   }, [search, tasks]);
 
+  useEffect(() => {
+    if (divRef.current) setListHeight(divRef.current.clientHeight);
+  }, []);
```
juhoeerola, line 63: This doesn't update the height if the page is resized, it only gets the initial height.

Not sure how you would listen to the changes though.

juhoeerola (changes requested):

juhoeerola (approved):
👍

PR #468, opened 2022-06-08
Fix overlapping title and clientShares in Planner Client Weights

AnttiVaarala:


frontend/src/pages/PlannerWeightsPage.module.scss
```diff
@@ -2,20 +2,19 @@
...
+    align-items: flex-start;
+    margin-right: 25px;
   }
 
   .clientShares {
     display: flex;
-    flex-grow: 1;
-    justify-content: flex-end;
-    margin-right: 25px;
+    margin: 0 34px 0 auto;
```
IidaKainu, line 17: Now that `margin-right: 25px` is added to the .title, this has some unnecessary margin
```suggestion
    margin-left: auto;
```

frontend/src/pages/PlannerWeightsPage.module.scss
```diff
@@ -2,20 +2,19 @@
...
   @extend .layout-row;
-  margin-left: 25px;
+  margin: 0 25px;
   align-items: flex-start;
 
   .title {
     @extend .layout-row;
     gap: 12px;
-    align-items: center;
+    align-items: flex-start;
```
IidaKainu, line 11: Noticed we can get rid of the extra div around the infoTooltip by setting the align as baseline
```suggestion
    align-items: baseline;
```

IidaKainu (changes requested):
Just a few small things

frontend/src/pages/PlannerWeightsPage.module.scss
AnttiVaarala: If there's not margin on the right, element overflows and part of it is hidden, the 34px is (padding 16px + border 1px)*2

IidaKainu: I can't seem to replicate the overflowing, would you mind showing how to? (on slack or here)

IidaKainu (approved):

IidaKainu (approved):

PR #469, opened 2022-06-08
Fix email-type field error overflowing

JimiRantanen:


frontend/src/components/forms/FormField.module.scss
```diff
@@ -25,4 +25,6 @@
   path {
     fill: $COLOR_RASPBERRY;
   }
+  overflow: hidden;
+  text-overflow: ellipsis;
```
juhoeerola, line 29:
```suggestion
  & > span {
    overflow: hidden;
    text-overflow: ellipsis;
  }
  & > svg {
    min-width: max-content;
  }
```

The overflow styles should work nicer on the span. With the current styles it's the span that overflows the div, not the text content, so the text overflow doesn't take effect.

Also noticed that the icon shrinks, it would be nice to keep constant size.

juhoeerola (changes requested):

IidaKainu (approved):

PR #470, opened 2022-06-08
Fix client overview email on the same line

AnttiVaarala:


frontend/src/components/Overview.module.scss
```diff
@@ -134,5 +134,11 @@
       @extend .layout-row;
       gap: 10px;
     }
+    a {
+      overflow-wrap: normal;
+      overflow: hidden;
+      text-overflow: ellipsis;
```
IidaKainu, line 140: Extending existing styles seems to also do the trick
```suggestion
      @extend .hide-text-ellipsis;
```

IidaKainu (approved):
Good catch 👍 Consider the comment before merging

PR #471, opened 2022-06-09
Fix leave roadmap

IidaKainu:
Fixes two things related to leaving a roadmap:
1. If the same roadmap was selected, unselect the roadmap
2. Refetch roadmaps (RoadmapSelectorWidget was not updated after
leaving a roadmap)

### DIFF
--- a/frontend/src/components/modals/LeaveRoadmapModal.tsx
+++ b/frontend/src/components/modals/LeaveRoadmapModal.tsx
@@ -14,6 +14,7 @@ import { ReactComponent as AlertIcon } from '../../icons/alert_icon.svg';
 import '../../shared.scss';
 import { RoleType } from '../../../../shared/types/customTypes';
 import { userInfoSelector } from '../../redux/user/selectors';
+import { chosenRoadmapIdSelector } from '../../redux/roadmaps/selectors';
 import { getType } from '../../utils/UserUtils';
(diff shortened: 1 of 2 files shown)


frontend/src/components/modals/LeaveRoadmapModal.tsx
```diff
@@ -52,6 +55,9 @@ export const LeaveRoadmapModal: Modal<ModalTypes.LEAVE_ROADMAP_MODAL> = ({
       if (res.payload?.message) return setErrorMessage(res.payload.message);
       return;
     }
+    if (currentRoadmapId === roadmapId)
+      dispatch(roadmapsActions.clearCurrentRoadmap());
+    refetchRoadmaps();
```
IidaKainu, line 60: If anyone knows how to use apiV2 in redux actions, refetching roadmaps could be moved to the leaveRoadmap-action. I just could not figure it out.

juhoeerola: I think this should do: https://redux-toolkit.js.org/rtk-query/usage/usage-without-react-hooks#performing-mutations

IidaKainu: Interesting; I thought I tried that out. Well, it works now 👍

PatricKangasmaki (approved):
Works well 👍

PR #474, opened 2022-06-10
Fix task value

juhoeerola:
The last commit tries to document the intended calculations. If the code disagrees with that, it's either bug in the code or the documentation isn't clear enough.


IidaKainu:
Ended up removing the total value from (milestone)ratingSummary, since the total depends on the amount of given ratings. We did not see where the data would be useful, since you would need to know the amount of raters too.

IidaKainu (approved):

PR #475, opened 2022-06-10
Separate already selected mappings from the mapping being selected

JimiRantanen:


frontend/src/components/IntegrationConfiguration.module.scss
```diff
@@ -56,7 +56,6 @@
 
 .mappings {
   @extend .layout-column;
-  gap: 20px;
```
juhoeerola, line 59: I think some gap would be good here.

Now the rows are right next to each other, but there is 5px margin on the top of the dropdown button which makes the small visual gap

frontend/src/components/IntegrationConfiguration.tsx
```diff
@@ -152,6 +152,11 @@ const MapStates: FC<{ configuration: IntegrationConfiguration }> = ({
             {selected && (
               <SetMapping name={selected.name} select={select(selected.id)} />
             )}
+            {statusMapping && statusMapping.length > 0 && (
```
juhoeerola, line 155: I'd prefer having the whole section under the same condition because multiple conditions may get out of sync and it's not always obvious the conditions are meant to be the same.

```tsx
{selected && (
  <>
    <div ... statusMappingLabel>
    </div>
    <SetMapping name={selected.name} select={select(selected.id)} />
  </>
)}
{columns && statusMapping && statusMapping.length > 0 && (
  <>
    <div ... statusMappingLabel>
    </div>
    {statusMappings.map(...
    ...
  </>
```

frontend/src/components/IntegrationConfiguration.module.scss
JimiRantanen: I liked how it looked like without the gap, but I added a 5px gap.

frontend/src/components/IntegrationConfiguration.tsx
```diff
@@ -152,24 +152,32 @@ const MapStates: FC<{ configuration: IntegrationConfiguration }> = ({
...
-                    remove={() =>
-                      removeMapping({ roadmapId, name, id, mappingId })
-                    }
-                  />
-                );
-              })}
+            {columns && statusMapping && statusMapping.length > 0 && (
+              <>
+                <div className={classes(css.statusMappingLabel)}>
+                  {t('Selected mappings')}
```
IidaKainu, line 158: Just a thought: would 'Saved mappings' fit the label better? We don't have much to communicate the mappings are being saved (I recall trying to find a save button when I first tried mapping)

In either case, let's add the label to English translations

IidaKainu (approved):
Looks great 👍

PR #478, opened 2022-06-14
Add route authorization

IidaKainu:
https://trello.com/c/7zZHCnY1/681-lis%C3%A4%C3%A4-route-authorization
RequireRole and requirePermission can be used in roadmap routes to require a certain role or permission in order to access the route. If the condition is not met, user is redirected to 404 page.

### DIFF
--- a/frontend/src/utils/requirelogin.tsx
+++ b/frontend/src/utils/requirelogin.tsx
@@ -1,10 +1,12 @@
 import { FC, ComponentType, useEffect, useState } from 'react';
 import { shallowEqual, useDispatch, useSelector } from 'react-redux';
 import { Redirect, useLocation, matchPath } from 'react-router-dom';
+import { hasPermission } from '../../../shared/utils/permission';
+import { Permission, RoleType } from '../../../shared/types/customTypes';
 import { StoreDispatchType } from '../redux';
 import { RootState } from '../redux/types';
 import { userActions } from '../redux/user';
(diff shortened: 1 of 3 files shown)


frontend/src/utils/requirelogin.tsx
```diff
@@ -71,3 +73,23 @@ export function requireVerifiedEmail<T>(
...
+};
+
+export const requirePermission = <T,>(
+  Component: ComponentType<T>,
+  required: Permission,
+): FC<T> => (props) => {
+  const role = useSelector(userRoleSelector, shallowEqual);
+  if (!hasPermission(role, required)) return <Redirect to={paths.notFound} />;
+  return <Component {...props} />;
+};
```
PatricKangasmaki, line 95: I'd rename these functions in a fashion that the comment wouldn't be needed. But I'll leave this up to you to decide since the comment does the same job anyways, I just prefer minimal comments with more explanatory naming.

PatricKangasmaki (changes requested):
Developer can still see the "Time estimation" link in the planner but clicking it redirects you to the 404 page. Otherwise working great 👍

IidaKainu:
Changed the names to more descriptive ones and removed mistakenly added requirement in time estimation -route. Thanks for the review 👍

PatricKangasmaki (approved):
Looking good 👍

PR #479, opened 2022-06-15
Add new modal and styling for task map info-tooltip

PatricKangasmaki:
Task map info matches Figma now. I also added a new icon and description for zooming in and out (it wasn't in Figma, opinions if it's needed are appreciated). 

I'd also like to hear opinions on these two approaches I took:
- I separated the data of `Overview` and `Actions` to their own file as they turned out to be really long (120 lines of code). Having them in the main modal file made it + 270 lines and it felt really cluttered.
- CSS file is split to sections by comments as it started to get hard to read.


frontend/src/components/TaskMap.module.scss
```diff
@@ -26,7 +26,39 @@ $ZOOM: var(--zoom, 1);
...
+
 .taskmapInfo {
   width: 27px;
   height: 27px;
+  cursor: pointer;
+}
+
+.infoDrawer :global .MuiDrawer-paper {
+  margin-top: 111px;
+  padding-bottom: 111px; // Margin top pushes the bottom of the drawer out of the picture without padding.
```
PatricKangasmaki, line 59: Latest push only changed this:

```
margin-top: 115px -> 111px;
padding-bottom: 115px -> 111px;
```

(if you look closely, the info-drawer is a few pixels lower than the background):
image

IidaKainu:
The width of the hr-elements (or the lines, did not yet look at the code) could be constricted, they are currently making the modal horizontally scrollable
image

IidaKainu:
Suggestion: can we add the info icon in the modal header (like the relations icon in planner)? It could connect the modal to the disappearing info button better...maybe
image

PatricKangasmaki:
Added both suggestions 👍

IidaKainu:
The positioning of the modal and task map control icons break a bit when navbar is wraps to two lines. However, task map is realistically usable only in a wider viewport, so no need to spend much time on this if it isn't easily fixable.
image

IidaKainu (changes requested):

IidaKainu (approved):
👍 

> I didn't touch this for now. I think it's better to use our limited resources for other things right now since this only happens on a viewport that's too narrow to use the whole page and even then, the breaking is so small that it doesn't really have any other effect than the small breaking of the correct height of the drawer (Not blocking anything etc.)

Sounds like a good decision

PR #482, opened 2022-06-21
Fix routing to 404 on roadmap page reload

IidaKainu:
Roadmap role can be undefined when reloading the page, which was not considered in the requirePermission or requireRole functions.


PatricKangasmaki (approved):
👍

PR #484, opened 2022-06-22
Planner/weighted values section

IidaKainu:
Mistakenly closed this PR


PatricKangasmaki (changes requested):
Looking good 👍 One thing that caught my eye is that in the weighted section, the order of the values go: `total value first, then average value` but then in the another section below, the order goes `average values first, then total value`. IMO it would be better for them to be in a same order for consistency.

image

IidaKainu:
> IMO it would be better for them to be in a same order for consistency.

I agree: added. Thanks for bringing this up, it bothered me a bit too.

PatricKangasmaki (approved):
👍

PR #485, opened 2022-06-22
Add content to SortableTask tooltip

IidaKainu:
- Total value and link to task overview -page is added to SortableTask tooltip, which is used in milestonesEditorPage
- customer dot is changed to business critical -icon if the value is more than 16
  - https://trello.com/c/ybjIZfEU/723-plan-milestones-bisness-kriittinen-ikoni


PatricKangasmaki (changes requested):
When you navigate to the task details page from the tooltip, should the back arrow on the task details take you back to the planner (similar function as the task relations drawer)? Currently it takes you to the task list page. Otherwise this is looking great 🔥

IidaKainu:
> When you navigate to the task details page from the tooltip, should the back arrow on the task details take you back to the planner (similar function as the task relations drawer)? Currently it takes you to the task list page.

That's such a good catch. Added 💯

IidaKainu:
Mistakenly closed the pr

PR #486, opened 2022-06-22
Fix data refetching

IidaKainu:


IidaKainu:
Interpreted the cause of the "roadmap users not updating on email change" -problem wrong. Actually the user's new email would not get updated in any project they are in. Should be fixed now.

PR #488, opened 2022-06-22
Fix add more colors for planner chart

AnttiVaarala:
If there's less colors than milestones they start to repeat themselves


frontend/src/components/PlannerChart.tsx
```diff
@@ -27,7 +27,14 @@ import { apiV2 } from '../api/api';
...
 const classes = classNames.bind(css);
 
-const colors = ['blue', 'red', 'green', 'orange'] as const;
+const colors = [
+  '#E24A4A',
+  '#53D4F0',
+  '#8CE434',
+  '#AA75EE',
+  '#FBD92A',
+  '#EA9128',
```
IidaKainu, line 36: Maybe these should be added to colors.module.scss

IidaKainu (changes requested):

frontend/src/components/PlannerChart.tsx
IidaKainu: And noting just to remember what was discussed in weekly: the values can be darkened

frontend/src/components/PlannerChart.module.scss
```diff
@@ -19,4 +21,38 @@
...
+:global(.recharts-cartesian-axis-tick-value) {
+  @extend .typography-pre-title;
+  fill: $COLOR_BLACK100;
+  letter-spacing: 0;
+}
+
+.title {
+  @extend .typography-body-bold;
+  display: flex;
+  flex-direction: row;
```
PatricKangasmaki, line 36: This could extend to the class `.layout-row` instead as it has been a default style in this project

frontend/src/components/PlannerChart.module.scss
```diff
@@ -19,4 +21,38 @@
...
+  align-items: flex-start;
+}
+
+.legendColumn {
+  border: solid 1px $COLOR_BLACK10;
+  border-radius: 10px;
+  padding: 15px;
+  display: flex;
+  flex-direction: column;
```
PatricKangasmaki, line 45: Same here, `@extend .layout-column`

frontend/src/components/PlannerChart.tsx
```diff
@@ -168,53 +197,60 @@ export const PlannerChart: FC<{
...
               id={vername}
               key={vername}
               name={vername}
               type="linear"
               dataKey={vername}
-              stroke={colors[index % colors.length]}
+              stroke={
+                colors[(versionKeyNames.length - 1 - index) % colors.length]
+              }
+              strokeWidth={2}
```
PatricKangasmaki, line 214: I'd change the strokeWidth to something like 1.5, this value 2 looks a bit too thick in my opinion. And those other two places as well where to strokeWidth was at 2

PatricKangasmaki:
Few suggestions above. Otherwise this is looking great 👍 I will not request changes now as I won't be able to re-review anymore after this but feel free to disagree with my points as well and apply what you feel like are good changes

IidaKainu:
Sorry Antti, I went ahead and added Patric's suggestions, because the summer vacation is waiting 🌞

IidaKainu:
Sorry Antti, I went ahead and added Patric's suggestions, because the summer vacation is waiting 🌞

IidaKainu (approved):

PR #491, opened 2022-06-23
Hide shares visualizations with only one client

IidaKainu:
One client always has 100% of "produced" values, so there is no need to show visualizations related to it


PatricKangasmaki (approved):
👍 👍

PR #492, opened 2022-06-23
Milestone value-shares visualization

IidaKainu:
Adds milestone value-shares visualization to planner
image
The width of the PercentageBar is calculated as following: total task value / max. total task value. 
- to determine the max. total value, the number of **all** clients * max rating value is used


PatricKangasmaki (approved):
Looking really good 👍

PR #493, opened 2022-06-25
Add keys to columns in task map info-modal

PatricKangasmaki:
The mapped divs were missing keys. I added a custom key to each object in `TaskMapInfoModalColumns.tsx`. Decided to go for this solution as you couldn't use the FC nor Translation elements directly as the key.

### DIFF
--- a/frontend/src/components/modals/modalparts/TaskMapInfoModalColumns.tsx
+++ b/frontend/src/components/modals/modalparts/TaskMapInfoModalColumns.tsx
@@ -1,5 +1,4 @@
 import { FC } from 'react';
-import { Trans } from 'react-i18next';
 import FullscreenIcon from '@mui/icons-material/Fullscreen';
 import RestartAltIcon from '@mui/icons-material/RestartAlt';
 import PanToolAltIcon from '@mui/icons-material/PanToolAlt';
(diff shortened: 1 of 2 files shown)


frontend/src/components/modals/modalparts/TaskMapInfoModalColumns.tsx
```diff
@@ -10,6 +10,7 @@ import ZoomInIcon from '@mui/icons-material/ZoomIn';
 export const overviewColumns: {
   Title: FC;
   columns: { Subtitle: FC; Description: FC; Icon?: FC }[];
+  key: string;
 }[] = [
   {
     Title: () => <Trans i18nKey="Task relations" />,
```
IidaKainu, line 16: The another option for the key is to use the title. Sorry I did not explain it coherently previously.
```suggestion
    title: "Task relations",
```
Then title should be used as following `<Trans i18nKey={title} />` instead of `<Title />`
For actionIcons, the Action could be used similarly

IidaKainu (approved):
A thought for the implementation below. I think the current one is fine too, so approving if you prefer to go with it

PatricKangasmaki:
Now I got what you meant 😄 Changed the modal content to strings and added them as keys 👍

PR #494, opened 2022-06-27
Add integration instructions in README

IidaKainu:
Otherwise complete, but JIRA key locations / additional instructions are missing.


PatricKangasmaki (approved):
Looking good 👍

PR #496, opened 2022-06-30
Add view settings to milestones-editor

IidaKainu:
Settings are saved to localStorage by setting string and current userId and roadmapId.

### DIFF
--- a/frontend/src/pages/MilestonesEditor.module.scss
+++ b/frontend/src/pages/MilestonesEditor.module.scss
@@ -1,5 +1,6 @@
 @import '../shared.scss';
 @import '../components/SortableTaskList.module.scss';
+@import '../components/forms/DropdownGreenFilled.module.scss';
 
 .tooltipIcon {
   width: 30px;
@@ -146,7 +147,6 @@
 .topbar {
   @extend .layout-row;
   min-height: 60px;
-  overflow-x: auto;
   align-items: center;
   border-bottom: 1px solid $COLOR_BLACK40;
   margin-left: 23px;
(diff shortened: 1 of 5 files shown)


frontend/src/pages/MilestonesEditor.module.scss
```diff
@@ -146,7 +147,6 @@
 .topbar {
   @extend .layout-row;
   min-height: 60px;
-  overflow-x: auto;
```
IidaKainu, line 149: This sadly brings back the double-scrollbar on the bottom of the page. 
With the line, the menu content would not pop over elements (instead locating itself inside the top bar, in a vertical scrollbox)

PatricKangasmaki: Well, I guess we can live with that. The double-scroll bar becomes visible only when your screen with is under 1035px, so it shouldn't be there too often and even when it's there, it's nothing too bad. If you'd like to fix it, you could for example make the navbar go on two lines (if possible) or hide certain not-so-important-things there, such as the ones in the image, when going under certain width thresholds. Or make the weighted shares bar narrower etc. But IMO it's okay to leave as it is as well

image

PatricKangasmaki (approved):
Looking and working great 👍 I'll approve this now and leave the double-scrollbar up to you if you have time to try to fix it or not

frontend/src/pages/MilestonesEditor.module.scss
IidaKainu: Good suggestions, but I'll leave it as is due to the time constraints

